PR #69, opened 2020-06-16
Support for custom reaction

misa-coder:
Now it supports custom reaction for the giveaways too.


NicoHd105:
So can a custom emoji be used for the reaction with that change?

misa-coder:
Yahh @NicoHd105 I tested in my bot and it works with both custom and discord emojis

misa-coder:
The main issue is actually in 75 line where androz use packet.d.emoji.name where giveaway.reaction is an id that's why it always return in custom emoji case.

NicoHd105:
Does the "giveawayreactionAdded" event works for you when you use a custom emoji?

misa-coder:
Yahh @NicoHd105 that's why I pull this request

NicoHd105:
👍Thanks

Issue #41, opened 2020-04-28
exemptMembers:


AndromCoders:
Hello,

adding in exemptMembers: (members) => !members.roles.some((r)=> r.name ==='Role name'),

to the .start options

it seems not to recognize the winner has the role or not the role.  the code still selects the winner without the role.

other than that all seems to be running

discord.js  Library version :v12.2

ShreyashKira:
Same issue I have in my bot

Androz2091:
Hello
Maybe it's because the member is not cached. What is the size of your server?

ShreyashKira:
Yeah, its almost 190 members. It is cached because discord.js says, 250 members.

TalonSilverStaff:
Same 563 members. 6 members on testing server.  Cached the members same issue. Seems like it might be 100 cache limit.

NicoHd105:
same

ShreyashKira:
Is someone finds any solution for this

Androz2091:
Can you install v4.2.0 and check if this has been fixed for you?

ShreyashKira:
Yeahh now fixed for me.

PR #72, opened 2020-06-24
Support exempt members value

misa-coder:
i tested it on my personal copy and it works fine.

Solve issue #41


misa-coder:
Yahh i never thought like that. I never use default manager option so :(
I use it with custom exemptMembers and it works fine.

misa-coder:
---i find a better way to execute this---

PR #73, opened 2020-06-27
Support exemptMembers

misa-coder:
i use this with too many giveaways and it shows no error. I m not able to find any case which can reproduce any bug.
I like if you help me to find.


Androz2091:
I don't think this is really needed. Once the owner has fixed the bug, the giveaway will end, right?

misa-coder:
Thats the main issue here you can't edit sqlite values directly. They need to have big brain to actually execute this. So having a try catch is better option.

Androz2091:
Hmm, I see. Yeah, once the function is saved in the file, it's difficult to edit it...

PR #76, opened 2020-06-30
Edit readme.md

misa-coder:
Delete giveaway do not filter this.giveaways so commands like edit, reroll and end still works. Using this will solve customDB issue


Androz2091:
This was already fixed by 2d42724.

Issue #77, opened 2020-06-30
terminal error spam after deleting giveaway manually or by command


powerflips:
After a user deletes a giveaway manually in the terminal appears this error below many times has already repeated in the terminal more than 1500 times the same error, I can not make this error stop appearing in the terminal, I'm using the package discord-giveaways and the bot project giveaways-bot Androz someone can help me?

`2020-06-29T17:56:09.244135+00:00 app[worker.1]: (node:4) UnhandledPromiseRejectionWarning: DiscordAPIError: Unknown Message
2020-06-29T17:56:09.244180+00:00 app[worker.1]:     at RequestHandler.execute (/app/node_modules/discord.js/src/rest/RequestHandler.js:170:25)
2020-06-29T17:56:09.244181+00:00 app[worker.1]:     at runMicrotasks (<anonymous>)
2020-06-29T17:56:09.244182+00:00 app[worker.1]:     at processTicksAndRejections (internal/process/task_queues.js:97:5)
2020-06-29T17:56:09.244274+00:00 app[worker.1]: (node:4) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 179)`

NicoHd105:
Bro,i already told you what you can do but as said when the deleted giveaway duration ends then it will get some errors again

NicoHd105:
Or you maybe can make it that the bot restarts after the code i provided you beacuse then no error will happen anymore, but that is not the best thing ecause restarts take about 1 minute...

powerflips:
The problem is that I don't host on a VPS I host on an instance where the dependencies are installed automatically, so I can't edit the package inside the node_modules folder

NicoHd105:
Oh ok, but then its not really possible to catch the error and do something with it when you cannot edit the package as long as androz doesnt add another version i guess...

powerflips:
Exactly, my friend, for me it would only be solved if he himself made this version in the package, but thank you very much, maybe in the future this will be solved

Androz2091:
This was fixed in [v4.2.0](https://github.com/Androz2091/discord-giveaways/releases/tags/v4.2.0).

PR #78, opened 2020-06-30
Delete giveaway once the message is not found

misa-coder:
Solve issue #77 as there is no point giveaway data for deleted giveaways.


src/Giveaway.js
```diff
@@ -255,6 +255,7 @@ class Giveaway extends EventEmitter {
                 message = await this.channel.fetchMessage(this.messageID).catch(() => {});
             }
             if (!message) {
+                this.manager.deleteGiveaway(this.messageID)
```
Androz2091, line 258:
```suggestion
                this.manager.deleteGiveaway(this.messageID);
```

PR #110, opened 2020-10-01
Update Giveaway.js

Shashank3736:
Current situation. If someone selects `this.botCanWin` true. Then only bots can win the giveaway. As filter will remove all the users. This change will take care of that.


Androz2091:
Indeed, you're right. 👍 Thank you

PR #115, opened 2020-10-24
Update Manager.js

parasop:


Lebyy (commented):
I don't think adding a space makes any difference?

Androz2091:
I think it's an error

Issue #122, opened 2020-11-17
TypeError: Cannot read property 'replace' of undefined


Ult1m1:
(node:24) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'replace' of undefined
at /home/container/node_modules/discord-giveaways/src/Giveaway.js:375:26
at runMicrotasks (<anonymous>)
at processTicksAndRejections (internal/process/task_queues.js:97:5)
(node:24) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)

NicoHd105:
Does this still happen/did you maybe remove _{winners}_ from your _winMessage_ key in your giveaway object?

Androz2091:
:tada: v4.3.0 is now released!

PR #45, opened 2020-05-05
"Last Chance to enter" embed

Shadowv7:

PR #128, opened 2020-12-07
v4.3.0

Androz2091:
This pull request contains changes for the next version. Feel free to test it and send reviews below, it would be helpful!
Another important question I'm asking myself: Should we remove support for Discord.js v11 entirely? I think we should. But don't hesitate to send a message here if you're using Discord.js v11!

## Added

* Embeds are now generated using methods in the GiveawaysManager class that can be overwritten! (c653901e482910feeeec73799e33dacf37f6b924)
* Eslint added, for a better code quality
* New `giveaway.extraData` property, useful to store data such as custom requirements, etc. (496e1916f1524a8fc4f1db71c91bc0cf0acf1d49)
* New `GiveawaysManager#giveawayDeleted` event. (496e1916f1524a8fc4f1db71c91bc0cf0acf1d49)
* `reroll()` now edits the giveaway embed instead of only sending a new message. **In progress - we should use the generate-embeds methods to generate this one instead of building it like that. (496e1916f1524a8fc4f1db71c91bc0cf0acf1d49)**

## Fixed

* `reroll()` method now uses the number of winners specified in options object (2896edba87c43eee73aba40004ce6c0c6baf0fa9)
* Fix `botsCanWin` check (da52a5408ae615d002f7d51712127d0e77f48eb7)
* Giveaway messages are now merged with the default ones to avoid issues like #122 (bedd3de2b320d69c589672826f3f4f82e1a24a01)

## Changes

* `giveaway#content` has been renamed to `giveaway#remainingText` (04ce6cd)


NicoHd105:
Seems all fine for me, I already had added all the changes through extending and so far everything worked fine. _But I'm not a pro coder so I cannot say that there are 0 issues though._ For_ the v11 support, I agree with you, especially because it does not get updated anymore and we are already _slowly_ heading to v13 so whoever still uses v11 should so or so sooner or later update to keep performance up and so it would not give you any benefit to still support it and if you remove it you could make the code a little more readable and maybe even improve performance because it does not have to check for the version anymore _but probably just saves a few ns_.

wolfier74:
Looks good

Lebyy:
When will the docs be updated?

src/Giveaway.js
```diff
@@ -289,11 +295,11 @@ class Giveaway extends EventEmitter {
         if (!reaction) return new Discord.Collection().array();
```
NicoHd105, line 295: I think we should do `return [];`, shouldn't we? Because is there any reason to create a empty discord collection and then return that as a array instead of just a empty JS array?

wolfier74:
When this will be merged?

ririko5834:
I think that support for v11 can be removed. Also will be this https://github.com/Androz2091/discord-giveaways/pull/45 added?

Snowflake107:
Removing `discord.js v11` is a good idea since `discord.js` itself dropped support for v11

src/Giveaway.js
Snowflake107: yeah, `new Discord.Collection().array();` is useless

wolfier74:
When this will be added?

Lebyy:
I suggest merging #45

PR #152, opened 2020-12-22
Fix MongoDB wrong link

GreepTheSheep:
The link redirects to mysql.js instead of mongodb.js


Androz2091:
Thank you, it has already been fixed by #146
(when opening a PR, it's better to choose the develop branch as the target :wink:)

PR #163, opened 2020-12-31
Prize placeholder for reroll

Jelcoo:
I added a opportunity to add the prize to a reroll message. If u want remove the readme change, since I wasn't sure about its necessity.


Nico105:
Question, are you still able to change the branch when you click on it? 
Because it would be better if you merge into "develop" and not directly into master.

PR #165, opened 2020-12-31
✨Add prize variable for reroll and add end method in examples

Jelcoo:
I added a opportunity to add the prize to a reroll message. If u want remove the readme change, since I wasn't sure about its necessity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,25 @@ client.on('message', (message) => {
 
 When you use the delete function, the giveaway data and the message of the giveaway are deleted. You cannot restore a giveaway once you have deleted it.
 
+### end a giveaway
+
+```js
+client.on('message', (message) => {
+    const args = message.content.slice(settings.prefix.length).trim().split(/ +/g);
+    const command = args.shift().toLowerCase();
+
+    if (command === 'end') {
+        let messageID = args[0];
+        client.giveawaysManager.end(messageID).then(() => {
+            message.channel.send('Success! Giveaway ended!');
+        })
+        .catch((err) => {
+            message.channel.send('No giveaway found for ' + messageID + ', please check and try again');
+        });
+    }
+});
+```
+
 ## 🇫🇷 Translation
 
 You can also pass a `messages` parameter for `start()` function, if you want to translate the bot text :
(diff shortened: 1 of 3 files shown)


Nico105:
_You did not have to close and open another pr, you just would have had to click on "Edit" which is next to "Open with" but nvm_
Everything seems fine

Jelcoo:
Yea this is my first fork/commit so I didnt know, but now i do

README.md
```diff
@@ -208,6 +208,25 @@ client.on('message', (message) => {
 
 When you use the delete function, the giveaway data and the message of the giveaway are deleted. You cannot restore a giveaway once you have deleted it.
 
+### end a giveaway
```
Nico105, line 211: End instead of end

README.md
```diff
@@ -208,6 +208,25 @@ client.on('message', (message) => {
...
+
+```js
+client.on('message', (message) => {
+    const args = message.content.slice(settings.prefix.length).trim().split(/ +/g);
+    const command = args.shift().toLowerCase();
+
+    if (command === 'end') {
+        let messageID = args[0];
```
Nico105, line 219: const

src/Giveaway.js
```diff
@@ -449,6 +449,7 @@ class Giveaway extends EventEmitter {
                 this.channel.send(options.messages.congrat
                     .replace('{winners}', formattedWinners)
                     .replace('{messageURL}', this.messageURL)
+                    .replace('{prize}', this.prize);
```
Nico105, line 452: Remove semicolon ";" and move the line one line above so it matches the `end()` function

Nico105 (changes requested):
Small Fixes - Suggestion

README.md
Jelcoo: What do u mean?

Nico105: let is unnecessary because you never reassign messageID, use const instead

Androz2091 (approved):
LGTM. Thank you 👍

Nico105 (approved):

Jelcoo:
Shit, I deleted the fork... Is that a problem?

Nico105:
> Shit, I deleted the fork... Is that a problem?

The changes are still here, only the source branch is know unknown.
https://stackoverflow.com/questions/50914358/how-do-i-recover-a-deleted-fork-in-github

PR #169, opened 2021-01-04
🐛 Serialize exemptMembers function to a string

Androz2091:
This serializes the exemptMembers function to a string, so it can be saved to the json file. I tested it and it works but maybe for some complex functions it doesn't. I will let this PR opened some days so if someone finds bugs they can report them here

### DIFF
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -96,6 +96,10 @@ class Giveaway extends EventEmitter {
         this.message = null;
     }
 
+    get exemptMembersFunction () {
+        return (this.options.exemptMembers && (typeof this.options.exemptMembers === 'function')) ? new Function(`return ${this.options.exemptMembers}`) : null;
+    }
+
     /**
      * The link to the giveaway message
      * @type {string}
@@ -168,9 +172,9 @@ class Giveaway extends EventEmitter {
      * @type {Function}
      */
     async exemptMembers(member) {
-        if (this.options.exemptMembers && typeof this.options.exemptMembers === 'function') {
+        if (this.exemptMembersFunction) {
             try {
-                const result = await this.options.exemptMembers(member);
+                const result = await this.exemptMembersFunction(member);
                 return result;
             } catch (error) {
                 console.error(error);
@@ -266,7 +270,7 @@ class Giveaway extends EventEmitter {
             embedColorEnd: this.options.embedColorEnd,
             botsCanWin: this.options.botsCanWin,
             exemptPermissions: this.options.exemptPermissions,
-            exemptMembers: this.options.exemptMembers,
+            exemptMembers: this.options.exemptMembers ? this.options.exemptMembers.toString() : undefined,
             reaction: this.options.reaction,
             requirements: this.requirements,
             winnerIDs: this.winnerIDs,
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -105,7 +105,11 @@ class Giveaway extends EventEmitter {
     }
 
     get exemptMembersFunction() {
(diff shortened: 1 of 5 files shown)


Androz2091:
Linked Stackoverflow issue: https://stackoverflow.com/a/45676430/11856499

src/Giveaway.js
```diff
@@ -96,6 +96,10 @@ class Giveaway extends EventEmitter {
         this.message = null;
     }
 
+    get exemptMembersFunction () {
+        return (this.options.exemptMembers && (typeof this.options.exemptMembers === 'function')) ? new Function(`return ${this.options.exemptMembers}`) : null;
```
Nico105, line 100:
```suggestion
        return (this.options.exemptMembers && (typeof this.options.exemptMembers === 'function')) ? (this.options.exemptMembers.toString().includes('function anonymous') ? this.options.exemptMembers : eval(this.options.exemptMembers)) : null;
```
```js
const test = '731583387235647498'
exemptMembers: (member) => !member.roles.cache.find(r => r.id === test)
```
Does not work because it would save "test" as "test" inside the DB and not as the assigned value

We can go around this "blockade" with
```js
const test = '\'731583387235647498\''
exemptMembers: new Function('member', `return !member.roles.cache.find(r => r.id === ${test})`)
```
but this works only with my suggested change

To summarize the problem: "Be able to use variable parameters"
_Lmk if there's any better solution_

I tested my change and it works fine
```js
exemptMembers: (member) => !member.roles.cache.find(r => r.id === '731583387235647498')
// or
const test = '\'731583387235647498\''
exemptMembers: new Function('member', `return !member.roles.cache.find(r => r.id === ${test})`)
```
both work with it

Nico105 (changes requested):

src/Giveaway.js
Nico105: Or we could do smth different
`exemptMembers: { filter: (member, arguments) => !member.user.username === arguments[0], arguments: ['Nico105']}`
or 
`exemptMembers: { filter: (member, arguments) => !member.user.username === arguments.name, arguments: { name: 'Nico105' } }`
I don't really have a preference but the 2 things in this comment are probably the most user friendly since the one I frist suggested makes everything to a string and like that you could accidently add a error to the filter

Nico105: @Androz2091 Tell me which one is best and I will change the rest of the code and make a pr to this branch 
or if you don't like my last 2 suggestions because it's a object and not a single function anymore then you can also accept my initial suggestion

Nico105: Or we could also just pass in `(member, extraData)` so people can store their arguments in `extraData` but that would maybe not be the best solution cause `extraData` should be for data that the features of the giveaway system itself do not need. So idk about that. As said tell me what you find best.

Androz2091 (commented):
@Nico105 I think that we should only serialize the function using serialize js. And yes, variables outside the function won't work, but anyway I want to keep the package easy to use. So I think that the best solution is to use simple serialization using serializejs

Nico105 (approved):

src/Giveaway.js
```diff
@@ -105,7 +105,11 @@ class Giveaway extends EventEmitter {
     }
 
     get exemptMembersFunction() {
-        return this.options.exemptMembers ? (this.options.exemptMembers.toString().includes('function anonymous') ? this.options.exemptMembers : eval(this.options.exemptMembers)) : null;
+        return this.options.exemptMembers
+            ? (typeof this.options.exemptMembers === 'string' && this.options.exemptMembers.includes('function anonymous'))
+                ? eval(`(${this.options.exemptMembers})`)
```
Nico105, line 110:
```js
console.log(eval("function anonymus() { return true }")) // undefined 
console.log(eval("(function anonymus() { return true })")) // function anonymus() { return true } // typeof === function
```
`bonusEntries` does not need this because there the whole array gets serialized and not the function

PR #182, opened 2021-01-14
✨ Send a message when a giveaway ended without a winner

SoftyMystic:
Hello, please rename the "noDeterminedWinner" if there's a better name to it, I just couldn't think of any better name 😅


Nico105:
https://discord-giveaways.js.org/GiveawaysManager.html#event:giveawayEnded
You know that this event exists?

Issue #108, opened 2020-09-25
[Suggestion] X Role has X Entries


Myzumi:
Please add a feature to allow like nitro boosters to get Double entries into a giveaway

NicoHd105:
It is planned

Shashank3736:
Best way to execute it is by making allowing us to pass functions like 
```js
entries: message.author.bot? 0 : 1
```
This will not bound the entries part just to the role but we can also do many things there.

ririko5834:
Can you add this?

Lebyy:
This would be a really good feature.

Nico105:
🎉 v4.5.0 is now released!

PR #187, opened 2021-01-17
✨ Bonus Entries

Nico105:
`console.log(userArray)` Giveaway.js Line 398 to see the array with the bonus entries

**Format**
```js
bonusEntries: [{ bonus: (member) => member.roles.cache.get('12345' ? 1 : null), cumulative: false }]
```

Thoughts, suggestions, improvements, roasts?
If someone has a better name for the property, I'm open.

### DIFF
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -455,7 +455,7 @@ class Giveaway extends EventEmitter {
             if (options.addTime) this.endAt = this.endAt + options.addTime;
             if (options.setEndTimestamp) this.endAt = options.setEndTimestamp;
             if (options.newMessages) this.messages = merge(this.messages, options.newMessages);
-            if (options.newBonusEntryFunctions) this.BonusEntryFunctions = options.BonusEntryFunctions;
+            if (options.newBonusEntryFunctions) this.bonusEntryFunctions = options.newBonusEntryFunctions;
             if (options.newExtraData) this.extraData = options.newExtraData;
             // Call the db method
             await this.manager.editGiveaway(this.messageID, this.data);
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -159,13 +159,13 @@ class Giveaway extends EventEmitter {
      * Members with any of these permissions won't be able to win a giveaway.
      * @type {Discord.PermissionResolvable[]}
      */
-    get exemptPermissions() {
+    get exemptPermissions () {
         return this.options.exemptPermissions || this.manager.options.default.exemptPermissions;
     }
 
     /**
      * The bonus entries for this giveaway
-     * @type {BonusEntry?}
+     * @type {BonusEntry[]?}
      */
     get bonusEntries () {
         const validBonusEntries = this.options.bonusEntries && this.options.bonusEntries.length && this.options.bonusEntries.every((v) => typeof v === 'object');
@@ -326,12 +326,12 @@ class Giveaway extends EventEmitter {
     async checkBonusEntries(user) {
         const member = this.channel.guild.member(user.id);
 
-        const check = async (obj) => {
-            const funct = obj[Object.keys(obj)[0]];
-            const entries = obj[Object.keys(obj)[1]];
-            if (typeof funct === 'function' && Number.isInteger(entries) && entries >= 1) {
+        const checkBonusEntry = async (obj) => {
+            const filter = obj[Object.keys(obj)[0]];
+            const bonus = obj[Object.keys(obj)[1]];
+            if (typeof filter === 'function' && Number.isInteger(bonus) && bonus >= 1) {
                 try {
-                    const result = await funct(member);
+                    const result = await filter(member);
                     if (result) return entries;
                 } catch (error) {
                     console.error(error);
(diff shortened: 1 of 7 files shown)


Androz2091:
I think we should use `serialize-javascript` to store the bonus entries in the JSON file and in the other databases.

```js
const serialize = require('serialize-javascript');

const bonusEntries = [
    {
        filter: (member) => member.roles.cache.some((r) => r.name === 'Booster'),
        bonus: 10
    }
]

const myStringBonusEntries = serialize(bonusEntries); // this is a simple string that can be stored anywhere

const retrievedBonusEntries = eval(`( ${myStringBonusEntries} )`);

// and retrievedBonusEntries is the same as bonusEntries!
```

I also added some commits to the PR, let me know if something was wrong

Nico105:
> I think we should use `serialize-javascript` to store the bonus entries in the JSON file and in the other databases.
> 
> ```js
> const serialize = require('serialize-javascript');
> 
> const bonusEntries = [
>     {
>         filter: (member) => member.roles.cache.some((r) => r.name === 'Booster'),
>         bonus: 10
>     }
> ]
> 
> const myStringBonusEntries = serialize(bonusEntries); // this is a simple string that can be stored anywhere
> 
> const retrievedBonusEntries = eval(`( ${myStringBonusEntries} )`);
> 
> // and retrievedBonusEntries is the same as bonusEntries!
> ```
> 
> I also added some commits to the PR, let me know if something was wrong

Added and tested with mongoose and .json

Nico105:
Or should the default manager bonusEntries and the giveawayObject bonus entries not be mergerd?
In the view of a abuser (=bot owner who gives himself 100 entries for every giveaway on every server his bot is in), it does not really matter because he can add it either way.
In terms of a honest bot owner it is more convient because you could say that for example if someone voted for your bot on top.gg then that member gets +1 entrie for every giveaway of the bot.
Or would that still be a bad thing? Idk

Lebyy:
When will this be implemented? Eagerly waiting for it

Nico105:
> When will this be implemented? Eagerly waiting for it

Within this month (February) maybe?
Androz is a busy man

Lebyy:
> > When will this be implemented? Eagerly waiting for it
> 
> Within this month (February) maybe?
> Androz is a busy man

I can understand his life xD he has so many packages to maintain plus his bot(s).

Nico105:
@Androz2091 I'm unsure, should we check giveaway object bonus entries AND `manager.default` bonus entries like for `exemptMembers`
Or should we check `manager.default` only when there are no giveaway object bonus entries like `exemptPermissions`?

https://github.com/Androz2091/discord-giveaways/pull/187/commits/01ad4009b271b40fdd3ec4b90543453c36733d73

Nico105:
And I need an answer to my #169 review so that I know how we want to handle "custom arguments" because that would affect the `bonusEntries` format in the giveaway object.

Nico105:
And should we maybe change the format to this: https://github.com/Androz2091/discord-giveaways/issues/108#issuecomment-699502945 ? 
Idk if it's more user friendly but then we wouldn't need bonus entry objects.

Nico105:
And should we maybe add a `addable: true` option to the object?
Because currently we get all the bonus entry numbers from the functions and choose the highest of them as the amount of bonus Entries.
But with that extra property a dev would be able to say that if a member fulfills 2 functions then those entries should get added up `filter 1 = true, filter 2 = true | =>| 2 entries + 4 entries = 6 entries` `checkBonusEntries(), entries[6]`

Androz2091:
Good idea, but I would use for `cumulative` instead of `addable`

src/Giveaway.js
```diff
@@ -308,6 +319,34 @@ class Giveaway extends EventEmitter {
...
+     * @returns {Promise<number|boolean>} The highest bonus entries the user should get or false
+     */
+    async checkBonusEntries(user) {
+        const member = this.channel.guild.member(user.id);
+        const entries = [];
+
+        if (this.bonusEntries.length) {
+            for (const obj of this.bonusEntries) {
+                const filter = obj[Object.keys(obj)[0]];
+                const bonus = obj[Object.keys(obj)[1]];
```
Androz2091, line 333: so here we should use properties instead of Object.keys

src/Giveaway.js
```diff
@@ -327,15 +366,44 @@ class Giveaway extends EventEmitter {
...
+
         const winners = [];
 
         for (const u of rolledWinners) {
             const isValidEntry = !winners.some((winner) => winner.id === u.id) && (await this.checkWinnerEntry(u));
             if (isValidEntry) winners.push(u);
             else {
                 // Find a new winner
-                for (const user of users.array()) {
+                for (const user of userArray || user.array()) {
```
Androz2091, line 406: user is not defined

typings/index.d.ts
```diff
@@ -41,6 +41,10 @@ declare module 'discord-giveaways' {
 
         public emit<K extends keyof GiveawaysManagerEvents>(event: K, ...args: GiveawaysManagerEvents[K]): boolean;
     }
+    interface BonusEntry {
+        filter (member: GuildMember): boolean|Promise<boolean>;
+        bonus: number;
```
Androz2091, line 46: So here (and everywhere else for logic), it would become:
```suggestion
        bonus (member: GuildMember): number|Promise<number>;
        cumulative: boolean;
```

src/Giveaway.js
Nico105: oops

src/Giveaway.js
Nico105: Changed 👍

Androz2091 (approved):
Thank you for your awesome contribution to the project! :grin:

PR #192, opened 2021-01-22
Update index.d.ts

ToastOnAStik:
Allow boolean type to `storage` option while creating the manager


Androz2091:
I don't see the case where it would be required. Can't you just set it to `undefined` or remove the option? 🤔

PR #195, opened 2021-01-22
✨ Bugfix/quickmongo | Manager starts too early/DB is not fast enough

Nico105:
Prevents Error
```
UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'forEach' of null
at GiveawayManagerWithOwnDatabase._init (C:\Privat\Coding\Bots\Discord\Bots\giveaways-bot-master\node_modules\discord-giveaways\src\Manager.js:454:22)
```
Because at the very first bootup (or when you delete the whole DB storage) there is nothing in the DB and the `giveawaysManager` starts faster than the `ready` event of the DB gets emitted and so `if ((await db.get('giveaways')) === null) await db.set('giveaways', []);` gets done too late and then manager gets an error because there is nothing in the DB.

Related to and uses solution of #153

If someone knows a better solution, please name it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ Other examples:
 
 - [MySQL example](https://github.com/Androz2091/discord-giveaways/blob/master/examples/custom-databases/mysql.js)
 - [MongoDB (Mongoose) example](https://github.com/Androz2091/discord-giveaways/blob/master/examples/custom-databases/mongoose.js)
-- [MongoDB (QuickMongo) example](https://github.com/Androz2091/discord-giveaways/blob/master/examples/custom-databases/quickmongo.js)
+- [MongoDB (QuickMongo) example](https://github.com/Androz2091/discord-giveaways/blob/master/examples/custom-databases/quickmongo.js) ⚠️ **Note**: Not recommended for high giveaway usage, use the `mongoose` example instead
 - [Enmap example](https://github.com/Androz2091/discord-giveaways/blob/master/examples/custom-databases/enmap.js)
 
 ```js
--- a/examples/custom-databases/quickmongo.js
+++ b/examples/custom-databases/quickmongo.js
@@ -69,7 +69,7 @@ client.giveawaysManager = manager;
 // DB is ready
 db.on('ready', async () => {
     if (!(await db.get('giveaways'))) await db.set('giveaways', []);
-    // Start the giveawaysMmanager only after the DB got checked to prevent an error
+    // Start the "giveawaysManager" only after the DB got checked to prevent an error
     client.giveawaysManager._init();
 });
 
--- a/examples/custom-databases/quickmongo.js
+++ b/examples/custom-databases/quickmongo.js
@@ -66,8 +66,10 @@ const manager = new GiveawayManagerWithOwnDatabase(client, {
 // We now have a giveawaysManager property to access the manager everywhere!
 client.giveawaysManager = manager;
 
+// DB is ready
 db.on('ready', async () => {
     if (!(await db.get('giveaways'))) await db.set('giveaways', []);
+    // Start the giveawaysMmanager only after the DB got checked to prevent an error
     client.giveawaysManager._init();
 });
 
--- a/examples/custom-databases/quickmongo.js
+++ b/examples/custom-databases/quickmongo.js
@@ -8,9 +8,6 @@ const Discord = require('discord.js'),
 // Load quickmongo
 const { Database } = require('quickmongo');
 const db = new Database('mongodb://localhost/giveaways');
-db.on('ready', async () => {
-    if ((await db.get('giveaways')) === null) await db.set('giveaways', []);
-});
 
 const { GiveawaysManager } = require('discord-giveaways');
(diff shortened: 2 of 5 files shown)


Androz2091 (approved):
Thank you 👍 
And there is no problem with the other examples?

Nico105:
> Thank you 👍
> And there is no problem with the other examples?

I might add a delay for `endedGiveawayLifeTime` removing because if there are a lot of ended giveaways in the `quick.db` .sqlite file then it spams `sqlite error: database is locked` because everything goes to fast, but it does not crash anything and it continues to remove so the error goes away after a time. 

`quickmongo` has the problem `VersionError: No matching document found for id "5fdc068ae36cd93e77ebe841" version 4791 modifiedPaths` which occurs on some functions but that is more a problem of the `quickmongo` package than of `discord-giveaways` and unfortunately their support channel did bot have an answer for the error when someone asked for it some time ago. I will try to figure smth out but as said it's mpre the databases fault then the giveawaysManagers.

Btw mongoose does not have the error of this pr because of the schema and model system it uses and it does not really use `await` as far I know.

Nico105:
_Ups was on phone 😂_

Nico105:
We have maybe found some double-roll issue with `quickmongo` and are going to test the fix, if it fixes it then I will add the fix in this pr.

ToledoSDL (approved):

Nico105:
My lastest commit uses `await` for the methods that edit or delete giveaways in the DB, that reduces `VersionError: No matching document found for id "5fdc068ae36cd93e77ebe841" version 4791 modifiedPaths "data"` errors because it has more time so fetch/edit. So in normal giveaway usage = not too much at the same time, then it helps.
But still if there are about 5 giveaways who end around the same time it can still occur, and also the `endedGiveawaysLifeTime` function will 90% trigger it because it does `forEach`, I still added `await` to that one too but it does not really help.
I guess as long as `quickmongo` does not change it's structure/process we cannot do more because they themself state that it's a beginner DB which is similar to `quick.db` and overall it's just more focused on little saving things and not really built for a whole giveaway system.

Snowflake107 (changes requested):
About
```js
.forEach(async (giveaway) => await this.deleteGiveaway(giveaway.messageID));
```

We should avoid using asynchronous `forEach` because we `cannot` resolve the promise for `.forEach` as it is a function that returns `nothing`. Instead of asynchronous `.forEach`, we should use `for ... of` loop.

```js
for (const giveaway of giveaways) {
    await this.deleteGiveaway(giveaway.messageID);
}
```

Nico105:
> About
> 
> ```js
> .forEach(async (giveaway) => await this.deleteGiveaway(giveaway.messageID));
> ```
> 
> We should avoid using asynchronous `forEach` because we `cannot` resolve the promise for `.forEach` as it is a function that returns `nothing`. Instead of asynchronous `.forEach`, we should use `for ... of` loop.
> 
> ```js
> for (const giveaway of giveaways) {
>     await this.deleteGiveaway(giveaway.messageID);
> }
> ```

Added 👍 _I didn't really learn promises yet so I didn't notice_
And because I assume you're basically the `quickmongo` owner, you also don't know any more regarding the `VersionError` I mentioned above? 
I already saw your [link](https://github.com/Automattic/mongoose/issues/9506) in the official support channel because someone reported the issue.
Because even with all the `await` added the error still occurs, is it really just because `quickmongo` isn't meant for giveaway workload? So nothing more we can do about it?

Androz2091:
@Snowflake107 do you have any additional information to share? otherwise I'll merge this PR and if we found a way to fix the rest of the issues we will commit it later

Nico105:
> > About
> > ```js
> > .forEach(async (giveaway) => await this.deleteGiveaway(giveaway.messageID));
> > ```
> > 
> > 
> > We should avoid using asynchronous `forEach` because we `cannot` resolve the promise for `.forEach` as it is a function that returns `nothing`. Instead of asynchronous `.forEach`, we should use `for ... of` loop.
> > ```js
> > for (const giveaway of giveaways) {
> >     await this.deleteGiveaway(giveaway.messageID);
> > }
> > ```
> 
> Added 👍 _I didn't really learn promises yet so I didn't notice_
> And because I assume you're basically the `quickmongo` owner, you also don't know any more regarding the `VersionError` I mentioned above?
> I already saw your [link](https://github.com/Automattic/mongoose/issues/9506) in the official support channel because someone reported the issue.
> 
> ```js
> raw.data = parsed.target ? Util.setData(key, Object.assign({}, raw.data), value) : value;
>     await raw.save()
>         .catch(e => {
>             return this.emit("error", e);
>         });
> ```
> 
> This is the part in the `quickmongo` code that is mentioned in the error path.
> Because even with all the `await` added the error still occurs, is it really just because `quickmongo` isn't meant for giveaway workload? So nothing more we can do about it?

https://stackoverflow.com/a/54795089/12985179

Androz2091 (approved):
LGTM. @Nico105 awaiting your approval.

Nico105:
I approve in text, (github won't let me approve my own pr) because of as explained reasons these changes are probably all we can do.

PR #203, opened 2021-02-01
🐛 Bugfix and improve roll and reroll function

### LINKED ISSUE #199: undefined value in winners
If you start a giveaway with more winners than participants at the end, an error occurs here because the array will fill up with an undefined value (L#329) and produce an error L#333: `TypeError: Cannot read property 'id' of undefined`. It's in Giveaway.js

Nico105:
The change will stop allowing the same winner(s) after **the first** (if the `winnerIDs` changes after the reroll and then there is another reroll, the original winners can win again) reroll because why would someone want that? _For rerference, look at #202_
And I optimized the rest of the roll function.
Fixes #199
Fixes "not saving edited giveaway data after reroll"
Thoughts?

### DIFF
--- a/src/Constants.js
+++ b/src/Constants.js
@@ -123,7 +123,7 @@ exports.defaultRerollOptions = {
     winnerCount: null,
     messages: {
         congrat: ':tada: New winner(s) : {winners}! Congratulations!\n{messageURL}',
-        error: 'No valid participations, no winners can be chosen!'
+        error: 'No valid participations, no new winner(s) can be chosen!'
     }
 };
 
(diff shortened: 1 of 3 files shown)


Androz2091:
This is a good idea (should be labeled "enhancement" because it is never said that this is the expected behavior, it is a new "feature"), I will review it as soon as possible 👍

Nico105:
LN 330 fixes #199, that's why someone made #205
And the changes in Manager.js fix another not-wanted-behaviour, that's why the bug label

Androz2091:
Oh all right, np 👍 
Yes it is better to open one PR per change

Androz2091 (approved):

PR #205, opened 2021-02-03
Update Giveaway.js

Tomato6966:
This fixxes when not enouhg members reacts


Androz2091:
Hello @Tomato6966, at this time the package chooses as many winners as possible. For example, if there are 10 winners and only 5 members who reacted, there will be 5 winners. Maybe we can add a `requiredWinnerCount` option so the bot will say that there are not enough entries?

Nico105:
Further discussion in #213 would be better.

PR #211, opened 2021-02-06
🏷️ Update data types

Nico105:
Thoughts? 
Things I updated wrong?
Things that should also get updated?

### DIFF
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -160,7 +160,7 @@ class Giveaway extends EventEmitter {
      * @type {Discord.PermissionResolvable[]}
      */
     get exemptPermissions() {
-        return this.options.exemptPermissions || this.manager.options.default.exemptPermissions;
+        return this.options.exemptPermissions.length ? this.options.exemptPermissions : this.manager.options.default.exemptPermissions;
     }
 
     /**
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -265,7 +265,7 @@ class Giveaway extends EventEmitter {
             embedColor: this.options.embedColor,
             embedColorEnd: this.options.embedColorEnd,
             botsCanWin: this.options.botsCanWin,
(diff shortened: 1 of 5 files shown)


Androz2091 (approved):

src/Giveaway.js
```diff
@@ -160,7 +160,7 @@ class Giveaway extends EventEmitter {
      * @type {Discord.PermissionResolvable[]}
      */
     get exemptPermissions() {
-        return this.options.exemptPermissions || this.manager.options.default.exemptPermissions;
+        return this.options.exemptPermissions.length ? this.options.exemptPermissions : this.manager.options.default.exemptPermissions;
```
Androz2091, line 163: I don't see the point of adding a `.length` here. If it is an empty array, it should be a valid value right? If the point of this `.length` is to check whether the value is an array, there is a proper way to do it using `Array.isArray()`.

Btw if the value is `undefined`, it will throw `cannot read property length of undefined` right?

Nico105: In the `data` getter I made `exemptMembers` to an empty array per default because the `manager.default` one is also an empty array per default. 
And that's why the `.length` is in place to check if  the array has any values because as said per default it would now be an empty array and `||` would not work with that.
Oh, but now I get it, before a restart would happen it would be still be `undefined` if it's not set because the DB data gets fetched only on startup.
So just add a `Array.isArray()` then right?

Androz2091: Yes, check if it is an array and if it is not, use the default manager option

Nico105: Or would it maybe be better to move what I added in the `data` getter to the `.start` fuction so that it would be a emtpy array per default on giveaway creation?

Androz2091: Hmm yes we can, when instantiating the Giveaway object (https://github.com/Androz2091/discord-giveaways/blob/develop/src/Manager.js#L200)

Nico105: Changed

Androz2091 (approved):
Thank you 👍

PR #212, opened 2021-02-08
Added maria db support exemple and find a bug for me in mysql exemple so fix it

TakeFy:


Nico105:
image
Bro you said this change is only needed if you use `mariaDB` didn't you?

TakeFy:
> image
> Bro this is not needed if you use normal mysql, if you change this then giveaways will not work after a restart with normal mysql.
> You said this change is only needed if you use `mariaDB` right? So why don't you have this change in the mariaDB example?

on my localhost i'm not using maria db but raw mysql and it's not working with JSON.parse because think about it you are creating JSON column but try to convert it in json

Nico105:
> > image
> > Bro this is not needed if you use normal mysql, if you change this then giveaways will not work after a restart with normal mysql.
> > You said this change is only needed if you use `mariaDB` right? So why don't you have this change in the mariaDB example?
> 
> on my localhost i'm not using maria db but raw mysql and it's not working with JSON.parse because think about it you are creating JSON column but try to convert it in json

It's not working when you are using `mysql2` to connect, if you use normal `mysql` then it works.

Nico105:
```js
const MySQL = require('mysql');
const giveaways = res.map((row) => JSON.parse(row.data));
```
Works
```js
const MySQL = require('mysql2');
const giveaways = res.map((row) => row.data);
```
Works



```js
const MySQL = require('mysql');
const giveaways = res.map((row) => row.data);
```
Not Works
```js
const MySQL = require('mysql2');
const giveaways = res.map((row) => JSON.parse(row.data));
```
Not Works

Nico105:
And so I think it would be better to call it `myslq2.js` example instead of `mariaDB` because we are using the `mysql2` package to connect and not the `mariaDB` package.

examples/custom-databases/mariaDB.js
```diff
@@ -0,0 +1,113 @@
...
+
+    // This function is called when the manager needs to get all the giveaway stored in the database.
+    async getAllGiveaways(){
+        return new Promise(function (resolve, reject) {
+            sql.query('SELECT `data` FROM `giveaways`', (err, res) => {
+                if (err) {
+                    console.error(err);
+                    return reject(err);
+                }
+                const giveaways = res.map((row) => JSON.parse(row.data));
```
Nico105, line 50:
```suggestion
                const giveaways = res.map((row) => row.data);
```

examples/custom-databases/mysql.js
```diff
@@ -46,7 +46,7 @@ const GiveawayManagerWithOwnDatabase = class extends GiveawaysManager {
                     console.error(err);
                     return reject(err);
                 }
-                const giveaways = res.map((row) => JSON.parse(row.data));
+                const giveaways = res.map((row) => row.data);
```
Nico105, line 49:
```suggestion
                const giveaways = res.map((row) => JSON.parse(row.data));
```

Nico105 (changes requested):
As mentioned

examples/custom-databases/mariaDB.js
```diff
@@ -0,0 +1,113 @@
+const Discord = require('discord.js'),
+    client = new Discord.Client(),
+    settings = {
+        prefix: 'g!',
+        token: 'Your Discord Token'
+    };
+
+// Load mysql
```
Nico105, line 8:
```suggestion
// Load mysql2
```

Nico105 (changes requested):

Nico105:
Everything's broken, I'll make a new pr.

Issue #107, opened 2020-09-20
Feature request


tunikakeks:
Please make it able to Stop a giveaway

NicoHd105:
So you mean that you can stop it and then you continue it some time later?

tunikakeks:
Yes, and that I can Stop it before it ends 

Example:

I will end the Giveaway when 100 Users on my Discord. If I have 100 Users on my Discord Server i would Stop it

Shashank3736:
Hmm, something like the opposite of all present giveaway.
I guess you mean a way to create giveaways that never end until we run `stop()` method. These type of giveaways have no reason to be created by the module as you can simply make an announcement where members can react and do an `eval` to get a random member from the reaction when you achieve 100 members. 

Still, you can achieve this from the current version.
- Create a giveaway with an expected time until you achieve that goal (100 members). If you have no idea about time then make it 1 year long.
- Now add a [`guildMemberAdd`](https://discord.js.org/#/docs/main/stable/class/Client?scrollTo=e-guildMemberAdd) event in your code. 
- Check if your server reached 100 members or not, if yes end the giveaway by using [`end`](https://discord-giveaways.js.org/GiveawaysManager.html#end) method, else return.
- If you reached the time limit before 100 members goal then extend the time by [`edit`](https://discord-giveaways.js.org/GiveawaysManager.html#edit) method.

By this way, you can easily achieve it without any hassle. But in the next release, I hope this type of things will become pretty easy to make.

But with your this feature request I think @Androz2091 should add one more function which we can customise on giveaway and manager creation. A function named as `condition` which will be checked before ending the giveaway.
- If `condition` function returns true the giveaway will be ended else it just edit the embed and say the preset message. This type of thing can help users to make a giveaway with minimum participants or anything which they need.

ririko5834:
Please add it

PR #213, opened 2021-02-10
✨ requiredParticipationCount, requiredWinnercount, noValidEndingInterval

Nico105:
Thoughts?
As far as I tested it, it works.
Better variable names are of course accepted.
Better implementation of course accepted.

### End Command
Currently `requiredParticipationCount` will only be trigger by auto-ending, an `end` command with the `end()` function ignores it.
But `requiredWinnercount` will always be triggered (= even with an `end` command)

_2nd sentence because otherwise we would have to move `requiredWinnercount` handling into the `roll()` function and then a `reroll` command would also be affected by `requiredWinnercount` and I don't think that that is really wanted. The only thing that we could do if we should do it like that (lmk your opinion) is adding an argument to the `end` and `reroll` function where people can set `useRequiredWinnercount` to true if they like but otherwise nothing will be affected._

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ You can pass an options object to customize the giveaways. Here is a list of the
 -   **options.hasGuildMembersIntent**: whether the bot has access to the GUILD_MEMBERS intent. It works without, but it will be faster with.
 -   **options.default.botsCanWin**: whether bots can win a giveaway.
 -   **options.default.exemptPermissions**: an array of discord permissions. Members who have at least one of these permissions will not be able to win a giveaway even if they react to it.
+-   **options.default.requiredParticipationCount**: the required number of participations a giveaway needs to have before it is able to end
+-   **options.default.noValidEndingInterval**: the amount of time a giveaway should get extended for if "requiredParticipationCount" or "requiredWinnerCount" get not reached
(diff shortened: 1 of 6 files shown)


Androz2091 (changes requested):
So, first of all thank you so much for your work :grin:
I *think* that we should remove `noValidEndingInterval`. Yes. I'm sorry. :grimacing:
I will do the changes, don't worry.

Nico105:
As a thought, we could scrap `requiredParticipationCount` and just use the ` requirements` option [that you already added to the giveaway data](https://github.com/Androz2091/discord-giveaways/blob/490e5c222e9518add406949244f0b1e938091246/src/Giveaway.js#L289)
= `endRequirements: (giveaway) => giveaway.message.reactions.cache.get(giveaway.reaction).users.cache.size - 1 >= 100`
#107

imranbarbhuiya:
Why it's still WIP?

Nico105:
> Why it's still WIP?

Waiting on Androz to tell the direction where this should go.

Nico105:
The future approach should be like mentioned in: https://github.com/Androz2091/discord-giveaways/pull/213#issuecomment-784138541
_I already started working on it a little, but it will take its time_

PR #216, opened 2021-02-18
🏷️ Add issue templates

Nico105:
It would be good to have issue templates because at the moment many issues don't have any structure/to less information is provided by the issue person and with templates we can tell people what should be provided if possible.
This might save a lot of unnecessary comments/questions by helpers in the issue itself.
Thoughts to my templates?
The issue and feature request template are 95% default from GitHub.
Suggestion for other templates?


Androz2091 (approved):
Thank you, there are perfect (except for the question one, the file name is missing a "e" :grin:)

PR #223, opened 2021-02-26
🐛 winnerCount checking more understandable

Lebyy:
Fixed the error where it said options.winnerCount is not a positive integer.

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -183,7 +183,7 @@ class GiveawaysManager extends EventEmitter {
             if (typeof options.prize !== 'string') {
                 return reject(`options.prize is not a string. (val=${options.prize})`);
             }
-            if (!Number.isInteger(options.winnerCount) || options.winnerCount > 0) {
+            if (!Number.isInteger(options.winnerCount) && options.winnerCount <= 0) {
                 return reject(`options.winnerCount is not a positive integer. (val=${options.winnerCount})`);
             }
             const giveaway = new Giveaway(this, {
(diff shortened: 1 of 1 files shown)


src/Manager.js
```diff
@@ -183,7 +183,7 @@ class GiveawaysManager extends EventEmitter {
             if (typeof options.prize !== 'string') {
                 return reject(`options.prize is not a string. (val=${options.prize})`);
             }
-            if (!Number.isInteger(options.winnerCount) || !options.winnerCount > 0) {
+            if (!Number.isInteger(options.winnerCount) || options.winnerCount > 0) {
```
Nico105, line 186:
```suggestion
            if (!Number.isInteger(options.winnerCount) || options.winnerCount < 1) {
```

src/Manager.js
```diff
@@ -183,7 +183,7 @@ class GiveawaysManager extends EventEmitter {
             if (typeof options.prize !== 'string') {
                 return reject(`options.prize is not a string. (val=${options.prize})`);
             }
-            if (!Number.isInteger(options.winnerCount) || !options.winnerCount > 0) {
+            if (!Number.isInteger(options.winnerCount) && options.winnerCount <= 0) {
```
Nico105, line 186:
```suggestion
            if (!Number.isInteger(options.winnerCount) || options.winnerCount < 1) {
```

Lebyy: Oh yes Number.isInteger is present thanks!

Lebyy: I think using an `&&` statement would be better here.

Lebyy: Since if the winner count is zero it will not throw an error.

Nico105: bro we are checking if is is not a number OR if the number is below 1
your change would check "only if it is not a number AND below 1 then return"

Lebyy: Ahhhhh Alrighty thanks for correcting me xD

PR #224, opened 2021-02-27
✨Drops feature

Lebyy:
I have implemented a drops feature in my bot and planning to maybe create a pr for it so it can be in discord-giveaways aswell.

https://user-images.githubusercontent.com/45114019/109376232-2c3f6f80-78e9-11eb-8a01-5ffe0065f524.mp4

Here is a video showcasing how it works

Basically a drop is when the user who reacts first wins.


Lebyy:
Confused if it should be added as a whole new instance like `client.giveawaysManager.startdrop` or just as an option in the giveaway like `isDrop: true`

Lebyy:
I think supplying it as an option is better.

Lebyy:
What do you think of it @Nico105?

src/Giveaway.js
```diff
@@ -461,6 +470,7 @@ class Giveaway extends EventEmitter {
             if (Array.isArray(options.newBonusEntries) && options.newBonusEntries.every((elem) => typeof elem === 'object'))
                 this.options.bonusEntries = options.newBonusEntries;
             if (options.newExtraData) this.extraData = options.newExtraData;
+            if(typeof options.newisDrop === 'boolean') this.isDrop = options.newisDrop;
```
Nico105, line 473:
```suggestion
            if (typeof options.newIsDrop === 'boolean') this.options.isDrop = options.newIsDrop;
```

src/Manager.js
```diff
@@ -186,6 +186,9 @@ class GiveawaysManager extends EventEmitter {
             if (!Number.isInteger(options.winnerCount) || options.winnerCount < 1) {
                 return reject(`options.winnerCount is not a positive integer. (val=${options.winnerCount})`);
             }
+            if (options.isDrop && typeof options.isDrop !== 'boolean') {
+                return reject(`options.isDrop is not boolean. (val=${options.isDrop})`);
```
Nico105, line 190:
```suggestion
                return reject(`options.isDrop is not a boolean. (val=${options.isDrop})`);
```

src/Manager.js
```diff
@@ -406,6 +410,9 @@ class GiveawaysManager extends EventEmitter {
                 await this.editGiveaway(giveaway.messageID, giveaway.data);
                 return;
             }
+            if (giveaway.options.isDrop == true && giveaway.message.reactions.cache.get(giveaway.reaction).count - 1 >= giveaway.winnerCount) {
```
Nico105, line 413:
```suggestion
            if (giveaway.options.isDrop && giveaway.message.reactions.cache.get(giveaway.reaction).count - 1 >= giveaway.winnerCount) {
```
Not needed because you already added that it makes sure that it only is a boolean

src/Manager.js
```diff
@@ -442,15 +449,14 @@ class GiveawaysManager extends EventEmitter {
...
-            giveaway.reaction || this.options.default.reaction
-        );
+        const reaction = message.reactions.cache.get(giveaway.reaction);
         if (!reaction) return;
         if (reaction.emoji.name !== packet.d.emoji.name) return;
         if (reaction.emoji.id && reaction.emoji.id !== packet.d.emoji.id) return;
         if (packet.t === 'MESSAGE_REACTION_ADD') {
             if (giveaway.ended) return this.emit('endedGiveawayReactionAdded', giveaway, member, reaction);
             this.emit('giveawayReactionAdded', giveaway, member, reaction);
+            if (giveaway.options.isDrop == true && reaction.count >= giveaway.winnerCount) this.end(giveaway.messageID).catch(() => {});
```
Nico105, line 459:
```suggestion
            if (giveaway.options.isDrop && reaction.count >= giveaway.winnerCount) this.end(giveaway.messageID).catch(() => {});
```
Not needed because you already added that it makes sure that it only is a boolean

typings/index.d.ts
```diff
@@ -153,6 +154,7 @@ declare module 'discord-giveaways' {
         newMessages?: Partial<GiveawaysMessages>;
         newBonusEntries?: BonusEntry[];
         newExtraData?: any;
+        newisDrop?: boolean;
```
Nico105, line 157:
```suggestion
        newIsDrop?: boolean;
```
CamelCase yk

typings/index.d.ts
```diff
@@ -180,5 +182,7 @@ declare module 'discord-giveaways' {
         embedColorEnd?: ColorResolvable;
         hostedBy?: string | null;
         extraData?: any;
+        lastChance?: LastChanceOptions;
+        newisDrop?: boolean;
```
Nico105, line 186:
```suggestion
        newIsDrop?: boolean;
```
CamelCase yk

Nico105 (changes requested):

src/Giveaway.js
Lebyy: Oops yes correct

src/Manager.js
Lebyy: Thanks xD got English exam tomorrow aswell LMAO

src/Manager.js
Lebyy: But what if someone set's it to false tho?

Lebyy: Oh yes true thanks for that correction.

src/Manager.js
Lebyy: Yep thanks for the correction.

typings/index.d.ts
Lebyy: Oops!

typings/index.d.ts
Lebyy: Oops xD

Lebyy (commented):
Thanks for the changes

Nico105 (approved):

ToledoSDL (approved):

Lebyy:
Should be case error oops.

README.md
```diff
@@ -189,9 +190,10 @@ client.on('message', (message) => {
 -   **options.newPrize**: the new prize.  
 -   **options.addTime**: the number of milliseconds to add to the giveaway duration.
 -   **options.setEndTimestamp**: the timestamp of the new end date (for example, for the giveaway to be ended in 1 hour, set it to `Date.now() + 60000`).
--   **options.newMessages**: the new giveaway messages
--   **options.newExtraData**: the new extra data value for the giveaway
+-   **options.newMessages**: the new giveaway messages.
+-   **options.newExtraData**: the new extra data value for the giveaway.
 -   **options.newBonusEntries**: the new BonusEntry objects (for example, to change the amount of entries).
+-   **options.newisDrop**: wheter to convert a drop to a normal giveaway or vice versa.
```
Nico105, line 196:
```suggestion
-   **options.newIsDrop**: wheter to convert a drop to a normal giveaway or vice versa.
```
Bro, [CamelCase](https://techterms.com/definition/camelcase) as said before, and you already approved `newIsDrop` in the constant, giveaway and types file...

Lebyy: Oh yes so sorry about that

Lebyy:
@Nico105 there is a HUGE problem, what if their bot restarts? And people react.

Lebyy:
So I think what I told you before about the Timeout method is more preferable.

Lebyy:
And I think bonus entries should be allowed for drops.

Nico105:
The drop will end when `reactions >= winnerCount` means that when the bot checks the giveaways 1st time after a restart and `reactions >= winnerCount` is true then it will end.

Lebyy:
> The drop will end when `reactions >= winnerCount` means that when the bot checks the giveaways 1st time after a restart and `reactions >= winnerCount` is true then it will end.

Ah alrighty

Lebyy:
This is necessary because all the members who react should win in a drop

Nico105:
> This is necessary because all the members who react should win in a drop

You mean because of `exemptMembers` or `exemptPermissions` ?
What if you want to like to make a role or permission exlusive drop...

Lebyy:
> > This is necessary because all the members who react should win in a drop
> 
> You mean because of `exemptMembers` or `exemptPermissions` ?
> What if you want to like to make a role or permission exlusive drop...
> You would force devs and bot users to let all ppl win which you shouldn't because this package should after all be customizable and not forced.
> If you want to let all ppl win you just don't set `exemptMembers` or `exemptPermission` for your giveaway. And the `manager.defaults` are defaults after all so they should apply for any giveaway
> So I don't see the point of doing it like you do.

No no, drops are like people who react before the number of winnercount is reached win

Lebyy:
> > This is necessary because all the members who react should win in a drop
> 
> You mean because of `exemptMembers` or `exemptPermissions` ?
> What if you want to like to make a role or permission exlusive drop...
> You would force devs and bot users to let all ppl win which you shouldn't because this package should after all be customizable and not forced.
> If you want to let all ppl win you just don't set `exemptMembers` or `exemptPermission` for your giveaway. And the `manager.defaults` are defaults after all so they should apply for any giveaway
> So I don't see the point of doing it like you do.

Also yes I need to add the validate member function

ToledoSDL (approved):

Nico105:
I wonder, should we maybe use `message.awaitReactions`?

Lebyy:
> I wonder, should we maybe use `message.awaitReactions`?
> Idk, is it faster/better than the discord data that gets received with the raw event?

But will it work after restart?

Nico105:
No, but we could just use the raw event for that again
And to avoid double endings we can just check in the raw event if the giveaway active time is lesser than the client uptime so it knows that `awaitReactions` is active for that specific giveaway
But that would only be worthly if `awaitReactions` is really better and idk about that, I will try to find out if it's maybe faster

Lebyy:
> No, but I would say we should just use the raw event for after-restart-giveaways again
> And to avoid double endings we can just check in the raw event if the giveaway active time is lesser than the client uptime so it knows that `awaitReactions` is active for that specific giveaway
> But that would only be worthly if `awaitReactions` is really better/faster and idk about that...

That can work 👍

Lebyy:
`awaitReactions` is way faster than the raw event.

Androz2091 (approved):
LGTM, thank you for your contribution

Issue #36, opened 2020-04-12
Support for more more winners?


menix1337:
Is it possible to adjust the giveaway if the description embed char reaches 255 > it then possibly skips writing names into the embed description?

And eventually if hitting max limit numbers in the say text (where it prints the winners names in) make multiple messages with more names if you have more winners than maybe 20?

All of this to adjust it being able to support more than 20ish winners per giveaway? Cause currently it just doesn't roll if you have above 20-25 winners, since it hits the description char limit from embeds? (Whcih could be avoided with if limit (255) > don't paste into description & just list multiple messages with winner names) to automatic adjust for a 'x' amount of winners? 

So if winners < 20 = it can update description to include winnner names.. If > 25, it doesn't upgrade the description with names - eventually just say (Winner(s) listed below) then print alot of rows with names.

Androz2091:
Yes, it will be added!

menix1337:
> Yes, it will be added!

Would be awesome! I know many who uses it, and with discords with over 17.000 members its annoying having to host 4 seperate giveaways for like 70-100 Winners! ^^

I am aware of at least 10 servers myself who Could have use of it ^^

menix1337:
Also is there a way to define each Winners discord ids? Lets say you have 100 winners (when this fix is in place)

You reckon its possible to award all Winners with a Discord rank? Provided by the bot using script - automaticly from their mentions or something.. ids?

Androz2091:
@menix1337 You should use the `giveawayEnded` event. You will have access to an array of winners. You'll be able to do something like that:
```js
giveawaysManager.on("giveawayEnded", (giveaway, winners) => {
    winners.forEach((winner) => winner.roles.add("role ID"));
});
```
Also, can you send me your Discord? It could help me to have an idea for how to fix your problems with large giveaways.

menix1337:
> @menix1337 You should use the `giveawayEnded` event. You will have access to an array of winners. You'll be able to do something like that:
> 
> ```js
> giveawaysManager.on("giveawayEnded", (giveaway, winners) => {
>     winners.forEach((winner) => winner.roles.add("role ID"));
> });
> ```
> 
> Also, can you send me your Discord server? It could help me to have an idea for how to fix your problems with large giveaways.

Sure, how would I go about to invite? I can easily toss up a test giveaway so you get the idea of the massive amount of people signing into it :D

Androz2091:
Yes, you can send the link here.

[...]

PR #225, opened 2021-02-28
🐛 Many winners embed support

Nico105:
image
So this should work. #36
The limits come from https://discordjs.guide/popular-topics/embeds.html#embed-limits
Any flaws in the code?


Nico105:
image
for testing

Lebyy:
> image
> So this should work. #36
> The limits come from https://discordjs.guide/popular-topics/embeds.html#embed-limits
> Any flaws in the code?

@Nico105 I think you might need to use the discord embed splitting because what if some users have a long username? It may hit the max chat count of embed descriptions.

Androz2091 (approved):

PR #232, opened 2021-03-05
Error on startup, fs throws an Invalid Argument TypeError

aanthr0:
image


Nico105:
Ahm did you accidently create a pull request instead of an issue? Because I don't see any commits from you.
And the error is probably because you have `storage: false` in your manager options right? 
You can just remove it to fix the error, we already removed it in the develop branch.

aanthr0:
There are no commits because I wasn't able to commit, and later I just forgot to check if I could.

Thanks for the reply though, cheers.

PR #234, opened 2021-03-09
🐛 Eris support

Nico105:
I just followed the [docs](https://abal.moe/Eris/docs) because I never touched eris 😆, so please tell when there are better methods to do things.
_I made this ~~draft~~ pr because of #229_

It would be good if someone with with real/good eris experience reviews this.

_I merged the dev branch at first, so because the base is now v13 some changes from dev get displayed in "files changed" too_

We should wait for: https://github.com/abalabahaha/eris/pull/1216
~~Currently using threads will break the pr because of bugs in the v9 pr (see link)~~


NotMarx:
@Nico105 you should fix `secondsBeforeLastChance` to `threshold` at [`src/Constants.js, line 136`](https://github.com/Nico105/discord-giveaways/blob/patch-3/src/Constants.js#L136) and also on the readme example for last chance.

src/Giveaway.js
```diff
@@ -408,32 +414,39 @@ class Giveaway extends EventEmitter {
...
+        if (!this.manager.libraryIsEris) {
+            const reactions = this.message.reactions.cache;
+            var reaction =
+                reactions.find((r) => r.emoji.name === Discord.Util.resolvePartialEmoji(this.reaction)?.name) ||
+                reactions.get(Discord.Util.resolvePartialEmoji(this.reaction)?.id);
+            if (!reaction) return [];
+        } else {
+            var reactionUsers = await this.message.getReaction(this.reaction);
+            if (!reactionUsers.length) return [];
+        }
```
Nico105, line 426: Should we maybe refrain from using var?
I doesn't make any problems, but its just maybe not "best practice"?
Originally I just made it like that so I don't have to make 2 let's on top, since 1 variable name for both doesn't make sense, at least to me.
Or are there any other solutions?

NotMarx: Just use const I guess..??

NotMarx:
In my opinion, it would be better if it can integrate with Eris in its own separate branch instead of doing this method.

Nico105:
> In my opinion, it would be better if it can integrate with Eris in its own separate branch instead of doing this method.

I already asked Androz about that and he said it would be harder to maintain.

src/Giveaway.js
Nico105: Cant use outside of the if, or hdym?

NotMarx:
> I already asked Androz about that and he said it would be harder to maintain.

In what way it would be hard to maintain? I mean, separate branches would mean that you don't actually really need to maintain both.

Androz2091:
> In what way it would be hard to maintain? I mean, separate branches would mean that you don't actually really need to maintain both at the same time.

Let's say we fix something on the discord.js branch, we will probably need to do the same for the eris branch but we can forget (at least from time to time...) and the branches will probably get out of sync quickly. And this will lead to a super difficult project in terms of maintenance. Also, we will not be able to push the eris version to NPM and thus keep good version management for eris users.

NotMarx:
> Let's say we fix something on the discord.js branch, we will probably need to do the same for the eris branch but we can forget (at least from time to time...) and the branches will probably get out of sync quickly. And this will lead to a super difficult project in terms of maintenance. Also, we will not be able to push the eris version to NPM and thus keep good version management for eris users.

Well, you're right there but we can just slightly ignore if the Eris branch wouldn't be as up-to-date as the discord.js branch for now as there are only a few people who use Eris and the library development itself has been slowed down these days so maintaining the Eris version wouldn't be a problem for now, at least. We might not able to push the Eris version to NPM but we can do eg. `npm install Androz2091/discord-giveaways#eris` and there won't be any problem with that. 

Not sure if this still sounds that getting a separate branch for the Eris version would be a great idea for now but it would be great if it does. And since I'm most likely an Eris dev, I could also contribute to the project for library code changes.

Also, even though making a separate branch for Eris itself isn't a good idea, using the current method right now (`this.libraryIsEris ? :`) is the same thing and this can also be very hard to maintain as well. 

Let me know if my opinions make sense or this wouldn't be possible and still hard to maintain. 😋

NotMarx (approved):
Tested everything including the newest pause/unpause features and no problems found at all. This seems to be working for me, at least.

src/Giveaway.js
```diff
@@ -534,30 +555,41 @@ class Giveaway extends EventEmitter {
...
+                                (this.message.channel.type !== 12 || this.message.channel.member || this.message.channel.permissionsOf(this.client.user.id).has('MANAGE_THREADS')) &&
+                                this.message.channel.permissionsOf(this.client.user.id).has(this.message.channel.threadMetadata.locked ? 'MANAGE_THREADS' : 'SEND_MESSAGES') &&
+                                this.message.channel.permissionsOf(this.client.user.id).has(this.message.channel.type === 12 ? 'USE_PRIVATE_THREADS' : 'USE_PUBLIC_THREADS')
+                            )
+                            : !this.message.channel.sendable && !this.message.channel.permissionsFor(this.client.user)?.has([
+                                this.message.channel.locked ? 'MANAGE_THREADS' : 'SEND_MESSAGES',
+                                this.message.channel.type === 'GUILD_PRIVATE_THREAD' ? 'USE_PRIVATE_THREADS' : 'USE_PUBLIC_THREADS',
+                            ]))
+                    ) ? this.libraryIsEris
+                            ? await this.message.channel.guild.channels.find(async (c) => [0, 5].includes(c.type) && (await c.getMessage(this.message.channel.parentID)))
```
Nico105, line 573: Is there no other way to get a message without providing the channel?

NotMarx: I don't think so.

Dragonizedpizza:
What if you used raw api and made it support all libraries

Nico105:
> What if you used raw api and made it support all libraries

"All libraries", if you look at the api site then you will only find discord.js and eris as official libraries for javascript. So we would have "all" covered already.

_The typescript libraries dont really count for us IMO since they have very low usage_

Also this would mean TREMENDOUS more work, for not much reward.

Nico105:
Concerning sharding (README).
Would it be
```js
return await Promise.all(client.shards.map(async () => await this.giveawaysManager.getAllGiveaways()))
```
for eris? (based on https://stackoverflow.com/a/37576787/12985179)

Dragonizedpizza:
Yes it would.

Dragonizedpizza:
> > What if you used raw api and made it support all libraries
> 
> "All libraries", if you look at the api site then you will only find discord.js and eris as official libraries for javascript. So we would have "all" covered already.
> 
> _The typescript libraries dont really count for us IMO since they have very low usage_
> 
> Also this would mean TREMENDOUS more work, for not much reward.

I mean... not **too** much work? You could just ask them to provide their websocket connection, and send to the api via some function wrappers.

You could make your own Message class,
```js
class Message {
  constructor(someParameters) {/** */}
  async edit(content) {
    const res = await fetch(`https://discord.com/api/v9/channels/${this.channelId}/messages/${this.id}`, {
      method: "PATCH",
      headers: { authorization: `Bot ${token}`, "Content-Type": "application/json" },
      body: JSON.stringify(content)
}).then(res => res.json());
 return res;
 }};
 ```
 What's even better about this is that when you upgrade to v13, those who don't want to upgrade from v12 can continue using v12 with discord-giveaways.

**Comment by Nico105** [comment]
> > > What if you used raw api and made it support all libraries
> > 
> > 
> > "All libraries", if you look at the api site then you will only find discord.js and eris as official libraries for javascript. So we would have "all" covered already.
> > _The typescript libraries dont really count for us IMO since they have very low usage_
> > Also this would mean TREMENDOUS more work, for not much reward.
> 
> I mean... not **too** much work? You could just ask them to provide their websocket connection, and send to the api via some function wrappers.
> 
> You could make your own Message class,
> 
> ```js
> class Message {
>   constructor(someParameters) {/** */}
>   async edit(content) {
>     const res = await fetch(`https://discord.com/api/v9/channels/${this.channelId}/messages/${this.id}`, {
>       method: "PATCH",
>       headers: { authorization: `Bot ${token}`, "Content-Type": "application/json" },
>       body: JSON.stringify(content)
> }).then(res => res.json());
>  return res;
>  }};
> ```
> 
> What's even better about this is that when you upgrade to v13, those who don't want to upgrade from v12 can continue using v12 with discord-giveaways.

wouldn't this be out of scope? 
Since the changes for eris weren't too many/serious, and as said we would now have the 2 official libraries covered already.
And regarding the backwards compatibility, we also dropped it for v11 eventually, and now Androz even said to dismiss the "transition phase support" from v12 to v13.  And eventually most ppl will switch regardless.
I just think it isn't really needed, to now communicate with the API ourselves.

**Comment by Dragonizedpizza** [comment]
> > > > What if you used raw api and made it support all libraries
> > > 
> > > 
> > > "All libraries", if you look at the api site then you will only find discord.js and eris as official libraries for javascript. So we would have "all" covered already.
> > > _The typescript libraries dont really count for us IMO since they have very low usage_
> > > Also this would mean TREMENDOUS more work, for not much reward.
> > 
> > 
> > I mean... not **too** much work? You could just ask them to provide their websocket connection, and send to the api via some function wrappers.
> > You could make your own Message class,
> > ```js
> > class Message {
> >   constructor(someParameters) {/** */}
> >   async edit(content) {
> >     const res = await fetch(`https://discord.com/api/v9/channels/${this.channelId}/messages/${this.id}`, {
> >       method: "PATCH",
> >       headers: { authorization: `Bot ${token}`, "Content-Type": "application/json" },
> >       body: JSON.stringify(content)
> > }).then(res => res.json());
> >  return res;
> >  }};
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What's even better about this is that when you upgrade to v13, those who don't want to upgrade from v12 can continue using v12 with discord-giveaways.
> 
> wouldn't this be out of scope?
> Since the changes for eris weren't too many/serious, and as said we would now have the 2 official libraries covered already.
> And regarding the backwards compatibility, we also dropped it for v11 eventually, and now Androz even said to dismiss the "transition phase support" from v12 to v13. And eventually most ppl will switch regardless.
> I just think it isn't really needed, to now communicate with the API ourselves.

Most people won't. Almost everyone knows discord.js@dev exists, yet people choose discord-buttons over it. Why? Because they don't have to change all their code. 

It wouldn't be too hard if you make your own message class, considering it then does all the hard work for you. 

Even if only discord.js & eris are the only libs in community resources, you'd end up helping alot of others who don't use discord.js/eris.

**Comment by Dragonizedpizza** [comment]
?

**Comment by Nico105** [comment]
> > > > > What if you used raw api and made it support all libraries
> > > > 
> > > > 
> > > > "All libraries", if you look at the api site then you will only find discord.js and eris as official libraries for javascript. So we would have "all" covered already.
> > > > _The typescript libraries dont really count for us IMO since they have very low usage_
> > > > Also this would mean TREMENDOUS more work, for not much reward.
> > > 
> > > 
> > > I mean... not **too** much work? You could just ask them to provide their websocket connection, and send to the api via some function wrappers.
> > > You could make your own Message class,
> > > ```js
> > > class Message {
> > >   constructor(someParameters) {/** */}
> > >   async edit(content) {
> > >     const res = await fetch(`https://discord.com/api/v9/channels/${this.channelId}/messages/${this.id}`, {
> > >       method: "PATCH",
> > >       headers: { authorization: `Bot ${token}`, "Content-Type": "application/json" },
> > >       body: JSON.stringify(content)
> > > }).then(res => res.json());
> > >  return res;
> > >  }};
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > What's even better about this is that when you upgrade to v13, those who don't want to upgrade from v12 can continue using v12 with discord-giveaways.
> > 
> > 
> > wouldn't this be out of scope?
> > Since the changes for eris weren't too many/serious, and as said we would now have the 2 official libraries covered already.
> > And regarding the backwards compatibility, we also dropped it for v11 eventually, and now Androz even said to dismiss the "transition phase support" from v12 to v13. And eventually most ppl will switch regardless.
> > I just think it isn't really needed, to now communicate with the API ourselves.
> 
> Most people won't. Almost everyone knows discord.js@dev exists, yet people choose discord-buttons over it. Why? Because they don't have to change all their code.
> 
> It wouldn't be too hard if you make your own message class, considering it then does all the hard work for you.
> 
> Even if only discord.js & eris are the only libs in community resources, you'd end up helping alot of others who don't use discord.js/eris.

won't? don't most ppl also use v12 now = changed from v11?
and what do you mean with "hard work"? As said (i think) this pr didn't change all too much.
and "who don't use discord.js/eris" are you talking about ppl who directly communicate with the API?

**Comment by Dragonizedpizza** [comment]
> > > > > > What if you used raw api and made it support all libraries
> > > > > 
> > > > > 
> > > > > "All libraries", if you look at the api site then you will only find discord.js and eris as official libraries for javascript. So we would have "all" covered already.
> > > > > _The typescript libraries dont really count for us IMO since they have very low usage_
> > > > > Also this would mean TREMENDOUS more work, for not much reward.
> > > > 
> > > > 
> > > > I mean... not **too** much work? You could just ask them to provide their websocket connection, and send to the api via some function wrappers.
> > > > You could make your own Message class,
> > > > ```js
> > > > class Message {
> > > >   constructor(someParameters) {/** */}
> > > >   async edit(content) {
> > > >     const res = await fetch(`https://discord.com/api/v9/channels/${this.channelId}/messages/${this.id}`, {
> > > >       method: "PATCH",
> > > >       headers: { authorization: `Bot ${token}`, "Content-Type": "application/json" },
> > > >       body: JSON.stringify(content)
> > > > }).then(res => res.json());
> > > >  return res;
> > > >  }};
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > What's even better about this is that when you upgrade to v13, those who don't want to upgrade from v12 can continue using v12 with discord-giveaways.
> > > 
> > > 
> > > wouldn't this be out of scope?
> > > Since the changes for eris weren't too many/serious, and as said we would now have the 2 official libraries covered already.
> > > And regarding the backwards compatibility, we also dropped it for v11 eventually, and now Androz even said to dismiss the "transition phase support" from v12 to v13. And eventually most ppl will switch regardless.
> > > I just think it isn't really needed, to now communicate with the API ourselves.
> > 
> > 
> > Most people won't. Almost everyone knows discord.js@dev exists, yet people choose discord-buttons over it. Why? Because they don't have to change all their code.
> > It wouldn't be too hard if you make your own message class, considering it then does all the hard work for you.
> > Even if only discord.js & eris are the only libs in community resources, you'd end up helping alot of others who don't use discord.js/eris.
> 
> _I oversaw that message_
> won't? don't most ppl also use v12 now = changed from v11?
> and what do you mean with "hard work"? As said (i think) this pr didn't change all too much/added completely different written code)
> and "who don't use discord.js/eris" are you talking about ppl who directly communicate with the API without libraries?

and those who are using lesser known libs, yes.
Hard work meaning, previously you would have had to send with a fetch or something, but a class makes it somewhat easy

File: src/Manager.js [hunk 3]
```diff
@@ -537,8 +552,30 @@ class GiveawaysManager extends EventEmitter {
...
+                    (await this.client.getRESTChannel(giveaway.channelId).catch(() => {})) ||
+                    (await this.client.getMessage(giveaway.channelId, giveaway.messageId).catch(() => {}))?.channel;
+                if (!channel?.id) continue;
+                if (!channel.name) {
+                    try {
+                        channel = require('../../eris/lib/structures/Channel').from(await this.client.requestHandler.request('GET', `/channels/${channel.id}`, true), this.client);
+                    } catch {
+                        channel = null;
+                    }
+                }
```
Nico105, line 570: To make this cleaner/better.
Temporarily make `client.options.restMode` true, use the rest method, and then return the option to its original state.
Would that be a good "workaround"?

Androz2091:
I think that this should not be added to the library. Eris is not really maintained anymore, so it will be hard to keep our lib supporting it. Thank you anyway for your huge work 🙏

PR #242, opened 2021-03-19
Fix error when winner undefined

megavirus2:
This error stop giveaway and don't edit embed of giveaway


megavirus2:
Ok, I checked develop branch, sorry

PR #243, opened 2021-03-19
Update GiveawaysManager.html

Dragonizedpizza:


Nico105:
Idk if this pr would change the website automatically when merged without clicking some button/if because of versioning it would even be good to merge it only for this one change

Explanation:
JSDoc can automatically use the comments in the code to generate a website
So the missing bracket your trying to fix has already been fixed in the referring comment in the dev branch (scroll down in manager.js)
Means when the dev branch gets merged and the new version is released androz just has to run a command and he gets the updated files

Androz2091:
The documentation is automatically generated (I don't even have to run a command myself) when something is pushed on the master branch.

Dragonizedpizza:
could you merge it then

Dragonizedpizza:
alr

PR #254, opened 2021-03-31
✨ Fetch all reaction users

Nico105:
Because we currently only fetch the first 100 users, but there could be more and we currently just don't include >100 users who participate = you have 0 chance to win when there are already 100 reactions
and so this also basically limits the number of winners to 99 (1 = bot)

this pr just fetches until there is no user left (= the fetched amount is less than 100 because 100 is the default fetch limit) because a discord user collection is sorted by id, so we can just use the `after` option to fetch more users
idk if this works fine tho, since where should I get the reactions, but I think it should? 😆

and because of rate limiting, most giveaways should stay under a couple of hundred reactions (<10 fetches), so it should not get limited right? / is there any proper way to deal with it in the case of occurring?

### DIFF
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -381,7 +381,14 @@ class Giveaway extends EventEmitter {
         const guild = this.channel.guild;
         // Fetch guild members
         if (this.manager.options.hasGuildMembersIntent) await guild.members.fetch();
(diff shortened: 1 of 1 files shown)


Nico105:
confirmed by #36

Androz2091 (approved):
Everything seems right! 👍

PR #258, opened 2021-04-06
✨ percent based bonus "entries"

### LINKED ISSUE #332: Bonus chance
**Is your feature request related to a problem? Please describe.**  

No

**Describe the solution you'd like**

It would be really good if you add an option for a bonus chance so that we can add a bonus chance for the users who vote for the bot on any bot listing site or used a referral code.
We can set the percentage of bonus chance like we use the edit, reroll option.

Nico105:
Mechanism based on https://stackoverflow.com/a/55671924/12985179

Note that this isn't totally bug tested by me, so it errors and bugs can occur.
_i dont understand whats really going on myself, the logging results just approximately matched up_

To activate percentages, look at the changed README

For testing:
```js
function weighted_random(options) {
    var i;

    var weights = [];

    for (i = 0; i < options.length; i++) weights[i] = options[i].weight + (weights[i - 1] || 0);

    var random = Math.random() * weights[weights.length - 1];

    for (i = 0; i < weights.length; i++) if (weights[i] > random) break;

    return options[i].item;
}

var users = [
    { item: 'user1', weight: 200 }, // 100 % more than everyone else = ~ double wins
    { item: 'user2', weight: 100 },
    { item: 'user3', weight: 100 },
    { item: 'user4', weight: 100 },
    { item: 'user5', weight: 100 },
];

// for (let i = 6; i<66; i++ ) users.push({ item: 'user' + i, weight: 1 })
console.log(users.length);
const results = [];
for (let i = 0; i < 100; i++) results.push(weighted_random(users));
users.forEach(obj => console.log(obj.item + ' ' + results.filter(v => v === obj.item).length));
```

if someone has a better mechanism feel free to suggest cause I don't even totally understand this one 😆
it behaves about accurate to the increased percentages, but there is the random factor after all so its not exactly double or tripple the chance, but approximately it is, I think

and also suggestions for the `usePercent` option are welcome cause I didn't think much about it and also I didn't really wanted to make it a seperate giveaway option/make `bonusEntries` a object, that's why i choose "last object in the array"

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,10 +281,14 @@ client.giveawaysManager.start(message.channel, {
         {
             bonus: (member) => member.roles.cache.some((r) => r.name === 'Nitro Boost') ? 2 : null,
             cumulative: false
-        }
+        },
+        { usePercent: false }
     ]
 })
 ```
+
+The `{ usePercent: false }` object should be the last object in the array, if you want to use it.
+If `usePercent` is set to true then the returned number of the `bonus` function of each object will get counted as %, means that if the function returns true for you then your win chance is 2% higher than that of other users with no increased win chance. 100% increase would mean about 2 times more winnings depending on the amount of other users.
 
 ⚠️ **Note**: If it should be customizable
 ```js
@@ -300,7 +304,8 @@ client.giveawaysManager.start(message.channel, {
         {   
             bonus: new Function('member', `return member.roles.cache.some((r) => r.name === \'${roleName}\') ? ${roleBonusEntries} : null`),
             cumulative: false 
-        }
+        },
+        { usePercent: false }
     ]
 })
 ```
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ client.giveawaysManager.start(message.channel, {
 ```
 
 The `{ usePercent: false }` object should be the last object in the array, if you want to use it.
-If `usePercent` is set to true then the returned number of the `bonus` function of each object will get counted as %, means that if the function returns true for you then your win chance is 2% higher than that of other users with no increased win chance. 100% increase would mean about 2 times more winnings depending on the amount of other users.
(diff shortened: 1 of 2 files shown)


Dragonizedpizza:
I didn't understand what percent based bonus entries mean.

Nico105:
> I didn't understand what percent based bonus entries mean.

It's not really entries anymore, its more like `increasedWinChance`, have you looked at the changed README ?, i explained it there.
It just wouldn't make sense to make something seperated from the original entries because you can only use entries or win chance not both together

Nico105:
Side thought: i guess we could make `usePercent` a option for the bonus entry objects, like `cumulative` and for those who don't use percentage we just set it to 0 in the rolling process?

= percentage is 0 but the user is 2 times in the array

_idk if 0 will cause problems/the user gets left out, since this is just a thought_

on the other hand, would this influence the results too much?

Lmk/i will maybe test it somewhen

xHyroM:
Math.random is not fully randomly.

Lebyy:
@Nico105 Here's a better mechanism to calculate the percentage based "entries"
```
var arrayShuffle = function(array) {
   for( var i = 0, length = array.length, swap = 0, temp = ''; i < length; i++ ){
      swap        = Math.floor(Math.random() * (i + 1));
      temp        = array[swap];
      array[swap] = array[i];
      array[i]    = temp;
   }
   return array;
};

var percentageChance = function(values, chances) {
   for ( var i = 0, pool = []; i < chances.length; i++ ) {
      for ( var i2 = 0; i2 < chances[i]; i2++ ) {
         pool.push(i);
      }
   }
   return values[arrayShuffle(pool)['0']];
};

//console.log(percentageChance(['msg', 'msg', 'msg'], [number%, number%, number%]));
console.log(percentageChance(['hi', 'test', 'bye'], [150, 100, 100]));
```

typhoon11:
What is the outcome of this PR then? I needed percent-based entries for the people who vote for my bot. Please add this fast

PR #275, opened 2021-05-01
🐛 fix: typings for deleteGiveaway and delete

Dev-CasperTheGhost:
* Minor: Return `boolean` instead of `void` 

This change was made because in the examples it says "Don't forget to return something" and to make TS happy 😅

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -308,7 +308,7 @@ class GiveawaysManager extends EventEmitter {
     /**
      * Delete a giveaway from the database
      * @param {Discord.Snowflake} messageID The message ID of the giveaway to delete
-     * @returns {Promise<void>}
+     * @returns {Promise<boolean>}
      */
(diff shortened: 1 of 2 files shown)


Nico105:
Hmm, I wonder if that should be done for the normal delete function as well, since it's also "public".
I assume it was made as "best pratice" in the DB examples to say "everything went well".

Dev-CasperTheGhost:
Possibly, let me know if I should change it too 😄

Nico105:
If you would be so kind

Dev-CasperTheGhost:
> If you would be so kind

Will do tomorrow since I’m about to get off for today

Dev-CasperTheGhost:
@Nico105 done 😄!

Nico105 (approved):

Nico105:
👍 Thank you

PR #294, opened 2021-05-28
✨ add ability to mention the giveaway host on end/reroll

idinium96:
Ping so that the giveaway host aware.

About the `toString()` method:
https://github.com/discordjs/discord.js/blob/c89bdd7566599a95a404b0f9e4b0828a866d0a71/src/structures/User.js#L301-L310


idinium96:
Oh.. in that case, I will just implement this for my usage only for now. thanks for the info. feel free to close this pull request :)

idinium96:
Or, I will add a check if `this.hostedBy` is not null/undefined

Androz2091:
#302 has been merged

PR #295, opened 2021-05-31
✨ style(*): Fix JSDoc spelling errors, and more.

WiseDevHelper:
This PR fixes a lot of the spelling errors in JSDoc comments and more.
ℹ️ This does not add any methods or parameters, making this a `SemVer: Patch` commit.

I'm open to any corrections and suggestions that may be suggested.

### DIFF
--- a/src/Constants.js
+++ b/src/Constants.js
@@ -19,7 +19,7 @@ const Discord = require('discord.js');
  * @property {string} [units.minutes='minutes'] The name of the "minutes" unit
  * @property {string} [units.hours='hours'] The name of the "hours" unit
  * @property {string} [units.days='days'] The name of the "days" unit
- * @property {Boolean} [units.pluralS='false'] Whether to force removing the "S" which marks the plural when the value is lower than 2
+ * @property {Boolean} [units.pluralS='false'] Whether to force the removal of "S" which marks the plural when the value is lower than two.
  */
 exports.GiveawayMessages = {};
 
@@ -39,7 +39,7 @@ exports.GiveawayMessages = {};
  * @property {Discord.ColorResolvable} [embedColorEnd] The color of the giveaway embed when it has ended.
  * @property {Discord.EmojiIdentifierResolvable} [reaction] The reaction to participate in the giveaway.
  * @property {GiveawayMessages} [messages] The giveaway messages.
- * @property {any} [extraData] The extra data value(s) for this giveaway.
+ * @property {any} [extraData] The extra data for this giveaway.
  * @property {LastChanceOptions} [lastChance] The options for the last chance system.
  */
 exports.GiveawayStartOptions = {};
@@ -169,7 +169,7 @@ exports.defaultRerollOptions = {
  * @property {number} [addTime] Number of milliseconds to add to the giveaway duration.
  * @property {number} [setEndTimestamp] The timestamp of the new end date.
  * @property {GiveawayMessages} [newMessages] The new giveaway messages.
- * @property {any} [newExtraData] The new extra data value(s) for this giveaway.
+ * @property {any} [newExtraData] The new extra data for this giveaway.
  * @property {BonusEntry[]} [newBonusEntries] The new BonusEntry objects.
  */
 exports.GiveawayEditOptions = {};
@@ -196,7 +196,7 @@ exports.GiveawayEditOptions = {};
  * @property {Discord.ColorResolvable} [embedColor] The color of the giveaway embed when it is running.
  * @property {Discord.ColorResolvable} [embedColorEnd] The color of the giveaway embed when it has ended.
  * @property {string?} [hostedBy] The mention of the user who hosts the giveaway.
- * @property {any} [extraData] The extra data value(s) for this giveaway.
+ * @property {any} [extraData] The extra data for this giveaway.
  * @property {LastChanceOptions} [lastChance] The options for the last chance system.
  */
 exports.GiveawayData = {};
--- a/src/Constants.js
+++ b/src/Constants.js
@@ -1,51 +1,51 @@
 const Discord = require('discord.js');
 
 /**
- * The Giveaway messages that are used to display the giveaway content
+ * The Giveaway messages that are used to display the giveaway content.
  * @typedef GiveawayMessages
  *
  * @property {string} [giveaway='@everyone\n\n🎉🎉 **GIVEAWAY** 🎉🎉'] Displayed above the giveaway embed when the giveaway is running.
- * @property {string} [giveawayEnded='@everyone\n\n🎉🎉 **GIVEAWAY ENDED** 🎉🎉'] Displayed above the giveaway embed when the giveaway is ended.
- * @property {string} [inviteToParticipate='React with 🎉 to participate!'] Displayed in the giveaway embed. Incite people to react to the giveaway.
- * @property {string} [timeRemaining='Time remaining: **{duration}**'] Displayed below inviteToParticipate in the giveaway embed. {duration} will be replaced automatically with the time remaining.
+ * @property {string} [giveawayEnded='@everyone\n\n🎉🎉 **GIVEAWAY ENDED** 🎉🎉'] Displayed above the giveaway embed when the giveaway has ended.
+ * @property {string} [inviteToParticipate='React with 🎉 to participate!'] Displayed in the giveaway embed. Invite people to react to the giveaway.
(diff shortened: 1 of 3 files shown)


src/Constants.js
```diff
@@ -1,51 +1,51 @@
...
- * @property {Discord.ColorResolvable} [embedColorEnd] The giveaway embeds color when they are ended
- * @property {Discord.EmojiIdentifierResolvable} [reaction] The reaction to participate to the giveaways
- * @property {GiveawayMessages} [messages] The giveaway messages
- * @property {any} [extraData] The extra data value for this giveaway
- * @property {LastChanceOptions} [lastChance] The last chance system options
+ * @property {Discord.ColorResolvable} [embedColor] The color of the giveaway embed when it is running.
+ * @property {Discord.ColorResolvable} [embedColorEnd] The color of the giveaway embed when it has ended.
+ * @property {Discord.EmojiIdentifierResolvable} [reaction] The reaction to participate in the giveaway.
+ * @property {GiveawayMessages} [messages] The giveaway messages.
+ * @property {any} [extraData] The extra data value(s) for this giveaway.
```
Nico105, line 42: I mean it can only have 1 value = singular. Setting it as an obj or an array does not mean that the property has multiple values. Or how do you interpret that?

src/Constants.js
```diff
@@ -78,13 +78,13 @@ exports.defaultGiveawayMessages = {
...
+ * The last chance options.
  * @typedef LastChanceOptions
  *
- * @property {boolean} [enabled] Whether the last chance system is enabled
- * @property {string} [content] The text of the embed when last chance is enabled
- * @property {number} [threshold] The number of ms after which the last chance system will be enabled
- * @property {string} [embedColor] The color of the embed when last chance is enabled
+ * @property {boolean} [enabled] Whether the last chance system is enabled.
+ * @property {string} [content] The text of the embed when the last chance system is enabled.
+ * @property {number} [threshold] The number of milliseconds before the giveaway ends which the last chance system will be enabled.
```
Nico105, line 86:
```suggestion
 * @property {number} [threshold] The number of milliseconds before the giveaway ends where the last chance system will be enabled.
```

src/Constants.js
```diff
@@ -94,26 +94,26 @@ exports.LastChanceOptions = {
...
+ * @property {Boolean} [default.botsCanWin=false] If bots can win giveaways.
  * @property {Discord.PermissionResolvable[]} [default.exemptPermissions=[]] Members with any of these permissions won't be able to win a giveaway.
  * @property {Function} [default.exemptMembers] Function to filter members. If true is returned, the member won't be able to win a giveaway.
- * @property {Discord.ColorResolvable} [default.embedColor='#FF0000'] The giveaways embed color when they are running
- * @property {Discord.ColorResolvable} [default.embedColorEnd='#000000'] The giveaways embed color when they are ended
- * @property {Discord.EmojiIdentifierResolvable} [default.reaction='🎉'] The reaction to participate in the giveaways
- * @property {LastChanceOptions} [default.lastChance] The last chance system parameters
+ * @property {Discord.ColorResolvable} [default.embedColor='#FF0000'] The color of the giveaway embed when it is running.
+ * @property {Discord.ColorResolvable} [default.embedColorEnd='#000000'] The color of the giveaway embed when it has ended.
+ * @property {Discord.EmojiIdentifierResolvable} [default.reaction='🎉'] The reaction to participate in the giveaway.
```
Nico105, line 110:
```suggestion
 * @property {Discord.ColorResolvable} [default.embedColor='#FF0000'] The color of the giveaway embeds when they are running.
 * @property {Discord.ColorResolvable} [default.embedColorEnd='#000000'] The color of the giveaway embeds when they have ended.
 * @property {Discord.EmojiIdentifierResolvable} [default.reaction='🎉'] The reaction to participate in giveaways.
```
Since this if for the manager options, it should stay in plural because it refers to all giveaways

src/Constants.js
```diff
@@ -138,18 +138,18 @@ exports.defaultManagerOptions = {
...
  * @typedef GiveawayRerollOptions
  *
- * @property {number?} [winnerCount=this.winnerCount] The number of winners to pick
- * @property {Object} [messages] The messages used in this method
- * @property {string} [messages.congrat=':tada: New winner(s): {winners}! Congratulations, you won **{prize}**!\n{messageURL}'] The message used if there are winners
- * @property {string} [messages.error='No valid participations, no new winner(s) can be chosen!'] The message used if no winner can be choosen
+ * @property {number?} [winnerCount=this.winnerCount] The number of winners to pick.
+ * @property {Object} [messages] The messages used in this method.
+ * @property {string} [messages.congrat=':tada: New winner(s): {winners}! Congratulations, you won **{prize}**!\n{messageURL}'] The message used if there are winners.
+ * @property {string} [messages.error='No valid participations, no new winner(s) can be chosen!'] The message used if no winners could be chosen.
```
Nico105, line 147:
```suggestion
 * @property {string} [messages.error='No valid participations, no new winner(s) can be chosen!'] The message used if no new winner(s) could be chosen.
```

src/Constants.js
```diff
@@ -161,42 +161,42 @@ exports.defaultRerollOptions = {
...
+ * @property {Discord.Snowflake} guildID The ID of the guild.
+ * @property {Discord.Snowflake?} [messageID] The ID of the message.
+ * @property {Discord.EmojiIdentifierResolvable} [reaction] The reaction to participate in the giveaway.
+ * @property {boolean} [botsCanWin] If bots can win the giveaway.
+ * @property {Discord.PermissionResolvable[]} [exemptPermissions] Members with any of these permissions won't be able to win the giveaway.
+ * @property {string?} [exemptMembers] Filter function to exempt members from winning the giveaway.
+ * @property {string} [bonusEntries] The array of BonusEntry objects for the giveaway.
+ * @property {Discord.ColorResolvable} [embedColor] The color of the giveaway embed when it is running.
+ * @property {Discord.ColorResolvable} [embedColorEnd] The color of the giveaway embed when it has ended.
+ * @property {string?} [hostedBy] The mention of user who hosts the giveaway.
```
Nico105, line 198:
```suggestion
 * @property {string?} [hostedBy] The mention of the user who hosts the giveaway.
```

Nico105 (changes requested):

src/Constants.js
WiseDevHelper: I think it should be `when` instead.

src/Constants.js
WiseDevHelper: Since this is the only place the user can add their extra data values, I think that `value(s)` would be better.

Nico105: but extraData can only have 1 value
Only the value of extraData can then have multiple values/properties if extraData is an array or obj
so "values/properties of value of extraData !== values of extraData" in my opinion

src/Constants.js
Nico105: Yeah, didn't had that in mind

src/Constants.js
Androz2091: Maybe we can... just remove the value x)
This would be the same: `The extra data for this giveaway (to store any additional things such as server requirements).`

Androz2091: and thank you for the PR, great ideas 👍

WiseDevHelper: > Maybe we can... just remove the value x)
> This would be the same: `The extra data for this giveaway (to store any additional things such as server requirements).`

Sounds great.

src/Constants.js
```diff
@@ -94,26 +94,26 @@ exports.LastChanceOptions = {
 };
 
 /**
- * The giveaways manager options
+ * The giveaways manager options.
  * @typedef GiveawaysManagerOptions
  *
  * @property {string} [storage='./giveaways.json'] The storage path for the giveaways.
  * @property {number} [updateCountdownEvery=5000] The giveaway update interval (in ms).
  * @property {number} [endedGiveawaysLifetime=null] The time (in ms) after which a ended giveaway should get deleted from the DB. ⚠ Giveaways deleted from the DB cannot get rerolled anymore!
```
Nico105, line 102:
```suggestion
 * @property {number} [updateCountdownEvery=5000] The giveaway update interval in milliseconds.
 * @property {number} [endedGiveawaysLifetime=null] The number of milliseconds after which ended giveaways should get deleted from the DB. ⚠ Giveaways deleted from the DB cannot get rerolled anymore!
```
Since you wrote it out for "threshold" too

Nico105 (changes requested):

Nico105 (approved):

Nico105:
I wonder, should we change the things that are the same as the comments you changed for JSDoc, in the README too, in this pr?

Nico105 (approved):
LGTM

PR #302, opened 2021-06-04
:sparkles: Access any giveaway property in giveaway messages

Nico105:
Since many people asked "how can i access ... in the ... message" and also "how can I make this ... message a embed".

embed:
```js
winMessage: { content: '', embed: {} }
```

access backend:
`'{code}' => regex => eval() => replace in string`

For an better explanation look at the 2 new ### in the README.
If it's not clear tho, just ask.

image
Extreme example

_EDIT: Since discord increased the embed description length to 4096, the ss is not that accurate anymore_


Nico105:
Makes pr's like #294 unnecessary

Nico105:
image
Btw I would also suggest, that if this gets merged, we change the default `messages.embedFooter` to `'{this.winnerCount} winner(s) • Powered by the discord-giveaways package'` (SS1) and the default `messages.winners` to `'Winner(s):'`. (SS2)

Because its a little bit weird for some people when they want to change (SS1)/remove it. because if they change `messages.winners` for (SS1) then they also change it for (SS2) (you could also say vice versa). It's just a little bit messed up.

And also it's a bit misleading when we call it `messages.embedFooter` even though it is NOT the whole footer.

Also we would save some code cause currently we make a weird process of formatting `messages.winners` for the different places.

Nico105:
Also depending on #330 , (if it gets taken over as the default) it might be a good idea to add smth like `editEmbedOnInterval: true` to the giveaway options, because if someone uses changing values of the giveaway then that person might want that it gets updated on the embed from time to time.

Androz2091 (approved):
Looks good. Indeed, it's better to allow more control to users about the messages. We should however keep the basic string. The goal of discord-giveaways is to stay **simple**. If users want something more complex, they should build it themselves.

Nico105:
> Looks good. Indeed, it's better to allow more control to users about the messages. We should however keep the basic string. The goal of discord-giveaways is to stay **simple**. If users want something more complex, they should build it themselves.

The basic string? Hdym that?
and isn't the package already beyond "simple" with bonus entries, etc?

Androz2091 (approved):
LGTM, thank you

PR #303, opened 2021-06-04
✨ end function, noWinnerMessage option

Nico105:
So some people want to send a message in the channel if there is no valid winner in the normal ending.
And ik and told people that i can be done with the event but since reroll also has the messages option and technically smth fast enough could come before the event I think it's alright to add this.

But so what confuses me is the original description of `giveaway.messages.noWinner` = `Sent in the channel if there is no valid winner for the giveaway.` since it suggest that it should have gotten sent in the channel, but i searched back and this message and description were added in v.3.0.0 but there never was code that sent it in the channel. Even before v.3.0.0 nothing got sent in the channel, so was idk was this just forgotten? Or was it never planned to be sent and the description just ended up like that because of stress?

Anyway since up until now no one complaint, I just changed the description of `noWinner` and made a separate option (noWinnerMessage) only for the end function which per default is null, so no one gets surprised.

But if someone (looking at androz) can explain the mystery of `giveaway.messages.noWinner`'s description, then that would be nice.


src/Manager.js
```diff
@@ -159,19 +159,20 @@ class GiveawaysManager extends EventEmitter {
 
     /**
      * Ends a giveaway. This method is automatically called when a giveaway ends.
-     * @param {Discord.Snowflake} messageId The message Id of the giveaway
+     * @param {Discord.Snowflake} messageID The message ID of the giveaway
+     * @param {string} [noWinnerMessage=null] The reroll options
```
Androz2091, line 163:
```suggestion
     * @param {string} [noWinnerMessage=null] Sent in the channel if there is no valid winner for the giveaway.
```

README.md
```diff
@@ -286,6 +286,8 @@ client.on('messageCreate', (message) => {
 });
 ```
 
+-   **noWinnerMessage**: Sent in the channel if there is no valid winner for the giveaway.
```
Androz2091, line 289:
```suggestion
- **noWinnerMessage**: Sent in the channel if there is no valid winner for the giveaway.
```

Androz2091 (approved):
Thank you 👍

PR #306, opened 2021-06-10
Schedule Giveaway

Aayansh549:
A Option to Schedule Giveaways like after 2 Days that Giveaway will start something like this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,10 +303,11 @@ client.giveawaysManager.start(message.channel, {
     time: 60000,
     winnerCount: 1,
     prize: 'Free Steam Key',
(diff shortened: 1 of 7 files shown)


Nico105:
Is this a feature request? Or an actual pr? Because you made a pr with a dependabot branch...

Aayansh549:
> Is this a feature request? Or an actual pr? Because you made a pr with a dependabot branch...

I doesn't know much about GitHub. That's why didn't get to Do that.

Nico105:
Ok so you want it as a feature.
I have smth in mind, I will work on it.

PR #307, opened 2021-06-10
✨ Add thumbnail support and use setTitle instead of setAuthor 

Ionys320:
The first change allows users to put an image with their giveaway, allowing it to be more attractive. It is an optional parameter of course, and using `.setThumbnail`.
The second change allows users to use custome emojis such as Steam logo with their prize.

VSC made auto-formats, sorry for the inconvenient.

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -61,7 +61,7 @@ class GiveawaysManager extends EventEmitter {
     generateMainEmbed(giveaway, lastChanceEnabled) {
         const embed = new Discord.MessageEmbed();
         embed
-            .setAuthor(giveaway.prize)
+            .setTitle(giveaway.prize)
             .setColor(lastChanceEnabled ? giveaway.lastChance.embedColor : giveaway.embedColor)
             .setFooter(`${giveaway.winnerCount} ${giveaway.messages.winners} • ${giveaway.messages.embedFooter}`)
             .setDescription(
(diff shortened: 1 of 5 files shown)


Nico105:
Custom emojis normally do not work in author?

And for the thumbnail option, I am not sure since it is not essential I think, and we have put the embed generating in functions where they can be extended for customization anyway. Also then we might just have to add the other remaining embed options as options too which may be too much to support per default. 
Tho I will wait for other opinions.

Ionys320:
> Custom emojis normally do not work in author?
I can confirm they are not working with author. I made few tests, and Discord send them as usual text when using auhor.

> I am not sure since it is not essential I think, and we have put the embed generating in functions where they can be extended for customization anyway. Also then we might just have to add the other remaining embed options as options too which may be too much to support per default.
Yeah, I understand. I would have the usage of thumbnails, especially because I do GWA for games, and put an image of the game without sending another message for this could be pretty nice, I think.

Thanks for your time!

src/Manager.js
```diff
@@ -197,11 +201,17 @@ class GiveawaysManager extends EventEmitter {
                 channelID: channel.id,
                 guildID: channel.guild.id,
                 prize: options.prize,
+                hostedBy: options.hostedBy ? options.hostedBy.toString() : null,
+                messages: options.messages,
+                reaction: options.reaction,
+                botsCanWin: options.botsCanWin,
+                exemptPermissions: Array.isArray(options.exemptPermissions) ? options.exemptPermissions : [],
```
Nico105, line 208:
```suggestion
```

Nico105:
👍 thank you for your contribution

PR #318, opened 2021-06-28
🐛 allowedMentions

Nico105:
allowedMentions.repliedUser not needed cause no reply. (mongoose schema)
`|| client.allowedMentions` as mentioned in the pr, not needed cause [discord does that automatically](https://github.com/discordjs/discord.js/blob/e37ef3af2151e556527ccd8e98ec531bdcc0ed70/src/structures/MessagePayload.js#L162-L165)
IMO option for the edit command is not needed because i think hardly anyone will change this option for 1 running giveaway.
I see it more of a option for server owners. = for all gaws in 1 server

tested.
thoughts?


Androz2091 (approved):
Thank you 👍

Androz2091 (approved):
LGTM 👍

ririko5834:
How to allow all mentions? Users are telling me that they are not pinged in the win message

Nico105:
> How to allow all mentions? Users are telling me that they are not pinged in the win message

this is not the right place to ask that btw
anyway, you should check your client settings first.
+ the users themselves could also have muted a server
otherwise you can use the `allowedMentions` option for the start function to overwrite your client defaults.
for values of the option, refer to the discord.js docs

Issue #330, opened 2021-07-20
Use new discord timestamps


DEVTomatoCake:
**Describe the solution you'd like**

Add an option to use the new [discord timestamp formatting](https://c.r74n.com/discord/formatting#Timestamps) in giveaways to display the remaining time.

This would make the message editing every few seconds unused.

Nico105:
You can already add it with the messages object option...

alima566:
Which option would that be? I don't see it anywhere in the link you provided.

Dragonizedpizza:
You could eliminate editing the message entirely with timestamp formatting.

Nico105:
> Which option would that be? I don't see it anywhere in the link you provided.

`<t:${(Date.now() + timeVariable) / 1000}:R>`
You just edit the message...

typhoon11:
> > Which option would that be? I don't see it anywhere in the link you provided.
> 
> `<t:${(Date.now() + timeVariable) / 1000}:R>`
> You just edit the specific message...

Don't forget to add `.toFixed(0)` or else it will give a decimal and then the formatting won't work.

Nico105:
Ah right right, or use math.round()

typhoon11:
But is there a way to stop the embed from editing, because even if I don't enter the `updateCountdownEvery`, it keeps on editing on its default value. So make an option if `updateCountdownEvery` is mentioned 0 it won't edit the embed

Androz2091:
Here is my take on the potential transition to the new Discord timestamps. Even though we won't have as much control over how the date is displayed as we used to (`in 24 hours and 6 minutes` would become `in a day`), I think using them for the next version of `discord-giveaways` would be very interesting.

The benefits are significant : 
* removing all spammy requests to Discord to edit messages
* displaying the end date in the current user's language
* better Discord integration (e.g., getting the full end date by hovering over the timestamp)

I would be curious to have the opinion of owners of bots using the library, feel free to comment.

alima566:
Removing the constant spammy requests to Discord to edit messages is a big plus. Currently removing the `updateCountdownEvery` does not work and still edits the message. Either removing this or setting this to 0, should stop the message from editing.

Also, if we have `lastChance` enabled, it should only edit the message once, once it reaches the threshold we specified.

Nico105:
We just remove the message.edit().
The `updateCountdownEvery` would still be needed, just as `checkGiveawaysEvery`.

Nico105:
[...]

PR #333, opened 2021-07-21
Make an option to assign updateCountdownEvery to zero

typhoon11:
When user's using the timestamp formatting that was introduced by Discord last month, users don't want the embeds to get edited, but looking on the current code it shows that if we don't mention the option `updateCountdownEvery` still it will take the default values and update it every 5 seconds, so I am suggesting this change that will allow user to set the value to 0 so that when the value of that field is zero it won't edit the embed.


Nico105:
You skipped a few thought/code steps but nvm.
I would say we wait for what exactly comes out of #330.

Nico105:
#358

Issue #324, opened 2021-07-09
Option to force reroll/end


Nico105:
**Is your feature request related to a problem? Please describe.**  

If giveaways get deleted from the db, either with a command or endedGiveawaysLifetime, but their message is still there, then currently they cannot get rerolled anymore.

**Describe the solution you'd like**

A `force: true` option for the reroll function.
Since the prize is in the title and winnerCount in the footer (generally) we can just attempt to get those 2, which are the most necessary, from the embed and then just create a temporary new instance of the giveaway class where we immediately call the reroll function.
<br/>
This is more like a "should we do this?" instead of a feature request btw.
Let me hear your opinions.

Nico105:
And it would probably make sense to add this option for the end function as well.

Nico105:
_see pr_

PR #315, opened 2021-06-26
🎉 Schedule Giveaways

Nico105:
Requested by some people.
Tested.
thoughts?

PR #336, opened 2021-07-27
[WIP] Discord.js v13

Nico105:
_Note: This not yet the new-version-pr for discord-giveaways, since it gets merged into the dev branch first_

⚠DO NOT UPDATE WHILE "OLD" GIVEAWAYS ARE STILL RUNNING!

## Changes

Support for/complete switch to the (coming) v13 version of discord.js
- New way of sending embeds
- ThreadChannel support

Other
- thingID to thingId => `messageID` to `messageId`
- use fs/promises & access() instead of exists() (=deprecated)
- edit() newLastChance option #346 

## PR's that could get merged / Things that need to be reviewed

- #318
- #315
- #302
  - https://github.com/Androz2091/discord-giveaways/pull/302#issuecomment-879280487
  - https://github.com/Androz2091/discord-giveaways/pull/302#issuecomment-886198338
- #234 ~~(Only when https://github.com/abalabahaha/eris/pull/1216 has no more relevant bugs concerning threads)~~
- #342 
- ~~#330 (Exact confirmation needed _IMO_) (If I get a YES, I will add it)~~ #358 
- #359 

Basically directed at Androz 😅
_Technically this pr can also get merged before those additional prs get merged. Order is up to Androz i guess._

## PR's where opinions would be needed (which maybe could get merged)

- #303 (Not directly v13 related but I mention it anyway)
- #224 (Wanted or not?)

Basically directed at Androz 😅

## Things that I will work on

- ~~#324 (I mention this because Idk if smth code related in v13 will be different, so it could also be that it will be just a pr for the normal dev branch)~~ #342 

## Questions

- ~~For intents should we use `Intents.FLAGS` or is providing strings fine enough?~~ everyone uses FLAGS apparently

- ~~https://github.com/Androz2091/discord-giveaways/issues/338#issuecomment-888352195~~

- ~~Should we throw an error if the `GUILD_MESSAGE_REACTIONS` intent is not provided?
Technically it is only necessary for 3 giveaway events (reaction add/remove).
So it does not have to be there.
But on the other hand people might get confused why they don't receive those 3 giveaway events.~~
Without the intent the reactions cache might be empty => now enforced

- Not really directly v13 relevant, but... Should the `giveaway.json` file that's currently in the package, which btw also gets published to npm (see package.json), get removed?
Cause I don't know the reason why it is there/gets published.
_If i oversaw something regarding that, pls lmk_

- ~~https://support-dev.discord.com/hc/en-us/articles/4404772028055-Message-Content-Access-Deprecation-for-Verified-Bots
Regarding that.
Should we switch the README command examples over to slash commands?
Or maybe remove the current handling without adding slash commands? Since for <100 guild bots it would still be valid but for verified bots not (unless intent, but I dont take that into consideration here, since it should be general)?
Or should we just leave it like it is now, since as mentioned, for <100 guild bots its still valid?
Lmk your thoughs.~~ #343
~~A team member in the discord.js server talked about that he will rewrite/adjust the slash example (guide), so we should wait for that for now anyway.~~ (= https://github.com/discordjs/guide/pull/786)

- the thread channel checker (if statement) you can see in the start command.
Should we put it inside a manager _function? Or a getter?
Because we already use it in start(), end() and reroll(). = 3 times
If "forcing" gets merged it would be +2 = 5.
also if "eris" gets merged it would get a bigger block.
Or do you think its okay like it is now?
EDIT: This might become irrelevant depending on dc.js v13.2. But we have to see.

## My thoughts

Did I forgot anything relevant concerning v13 features?

Reviews/opinions would be much appreciated.

## Status

- [x] This PR introduces some breaking changes.


zmkq:
there is a bug that let MessageID `undefined`

Nico105:
> there is a bug that let MessageID `undefined`

Everything was changed from thingID to thingId, to match discord.js v13

README.md
```diff
@@ -31,9 +31,18 @@ You can read this example bot on GitHub: [giveaways-bot](https://github.com/Andr
 
 ### Launch of the module
 
+Required Discord Intents: `GUILDS`, `GUILD_MESSAGES`, `GUILD_MESSAGE_REACTIONS`.
```
Nico105, line 34: Should the first 2 stay there?
Cause technically they would not be needed _I think_
But without them you also won't get to issue a command in your bot, in the first place.
But is that relevant for us?

lmk

Dragonizedpizza: its probably going to be required on like april

Nico105: Yeah, but I meant it like the first 2 aren't needed to start a gaw.
Which now proved to be wrong cause at least "GUILDS" is needed to get a channel from the clients cache.
So I will just leave it like that I guess.

PR #340, opened 2021-07-29
Add typing for giveawayDeleted event

NotMarx:
This event exists but has not been typed yet.

[Source](https://github.com/Androz2091/discord-giveaways/blob/e0ee06390ce2a929df64e9bbaf5a43914d94d559/src/Manager.js#L358)


Nico105:
👍

PR #342, opened 2021-08-01
🐛 Force reroll/ending + resume deleted giveaways

Nico105:
forcing:
```js
/**
 * The options for a forced ending or reroll of a giveaway.
 * The manager attempts to get "giveaway.messages" and "giveaway.embedColor|giveaway.embedColorEnd" (depending on "end or reroll"), from the giveaway that was started closest to the giveaway that is forced (in the same server).
 * @typedef GiveawayForceOptions
 *
 * @property {boolean} [force=false] If the ending or reroll should be forced, when the giveaway is not found.
 * @property {Discord.TextChannel|Discord.NewsChannel|Discord.ThreadChannel} channel The channel in which the giveaway that is forced is.
 * @property {boolean} [saveGiveawayInDatabase=true] If the giveaway that is forced should be saved in the runtime giveaway array and the database.
 * @property {number} [closestGiveawayThreshold=86400000] How big the "giveaway.startAt" difference to the closest giveaway can be, before the closest giveaway is considered invalid. Default is 1 day. If there is no valid closest giveaway then "giveaway.messages" and "giveaway.embedColor|giveaway.embedColorEnd" will have default values (Set this property to "0" to always use default messages). 
 */
```

resuming:
```js
manager.start(message.channel, {
    messageID: "12312312312311",
    prize: ''
    winnerCount: 1
    time: 100000
})
```

thoughts?

depending if #302 gets merged first, I would need to add a little code to the "auto winnerCount from footer" code.


src/Constants.js
```diff
@@ -151,11 +152,28 @@ exports.GiveawaysManagerOptions = {
...
+ * @property {boolean} [force=false] If the ending or reroll should be forced, when the giveaway is not found.
+ * @property {Discord.TextChannel|Discord.NewsChannel|Discord.ThreadChannel} channel The channel in which the giveaway that is forced is.
+ * @property {boolean} [saveGiveawayInDatabase=true] If the giveaway that is forced should be saved in the runtime giveaway array and the database.
+ * @property {number} [closestGiveawayThreshold=86400000] How big the "giveaway.startAt" difference to the closest giveaway can be, before the closest giveaway is considered invalid. Default is 1 day. If there is no valid closest giveaway then "giveaway.messages" and "giveaway.embedColor|giveaway.embedColorEnd" will have default values (Set this property to "0" to always use default messages). 
+ */
+exports.GiveawayForceOptions = {
+    force: false,
+    channel: null,
+    saveGiveawayInDatabase: true,
+    closestGiveawayThreshold: 86400000
```
Nico105, line 169: LMK your thoughts about this 1day default threshold.
I chose this cause giveaways are not all to frequent.
But you could say since thats the case, we could also make the default even bigger?
Or you guys think this default is fine?

Androz2091 (commented):
This is, for me, something that should definitely not be supported. The PR is great and we can keep it in mind in case there is a problem with the database of some users, but this is an edge-case that adds complexity to the module without a real interest. So I'm sorry for your PR, thank you anyway (it was great btw) 😅

Androz2091:
I am closing the PR but feel free to comment if you really need this feature

PR #344, opened 2021-08-02
✨ Apache CouchDB - Nano example

Nico105:
_I have nothing better to do_
Tested


Androz2091 (approved):
Not tested but looks good. Thank you. 👍

PR #358, opened 2021-08-15
✨ Use discord timestamps

Nico105:
Tested.

- current message: `drawing: 'Drawing: {timestamp}'`
lmk if something else looks (=name) better

- I currently just added the new functionality to the generateEmbed method, but maybe it should become a getter?

- should the README screenshots get adjusted because of this? Or can we leave them be?


ConnorDoesDev:
has this been merged? would love to see this in action.

Nico105:
> has this been merged? would love to see this in action.

It hasn't been merged, otherwise you wouldn't see it.
And technically you can already install this branch and try it out if you want.

ConnorDoesDev:
hi! I may be just too dumb to use this, but after removing `timeRemaining` from the messages object and adding `drawing` to it, it still uses the timeRemaining thing.

Nico105:
> hi! I may be just too dumb to use this, but after removing `timeRemaining` from the messages object and adding `drawing` to it, it still uses the timeRemaining thing.

I mean did you actually install this branch?
`npm i Nico105/discord-giveaways#patch-12` ?

zmkq:
i think the two choices (the current timer and the discord timestamps) would be better if there 2 options,
**Example** :
```js
timeRemaining: `Time Remaining {time}!`,
timeStampDiscord: `{drawing}`
```

Nico105:
> i think the two choices (the current timer and the discord timestamps) would be better if there 2 options,
> **Example** :
> 
> ```js
> timeRemaining: `Time Remaining {time}!`,
> timeStampDiscord: `{drawing}`
> ```

It is pretty much set in stone that we make a full switch.
The benefits performance wise are better for the general public.

Tho if #302 gets merged (pretty likely) then you could still use the old format if you really want.

Androz2091 (approved):
Awesome, thank you

PR #359, opened 2021-08-16
🐛 slash command examples

Nico105:
_I haven't really used slash commands yet, so don't roast me if smth is totally wrong 😆_

- should we make all `.catch` empheral ?

- Is it fine to write the `data` out?
Since we could remove that and for registering just do `delete command.run`.
= "save 1 object"
But that might confuse some people.

- Is `interaction.client.giveawaysManager` fine?
or should we provide the client as the 2nd command argument? since the above format could seem rather long.


Dragonizedpizza (changes requested):
I think the snake case in the reroll command is unnecessary, considering that this is for the users to view, and Id looks weird. Also, the options' names must be lowercase (Discord API requires so), and I feel like the arrow function `run: async (interaction) => {` is unnecessary, when you could just do `async run(interaction) {`

Nico105:
> I think the snake case in the reroll command is unnecessary, considering that this is for the users to view, and Id looks weird. Also, the options' names must be lowercase (Discord API requires so), and I feel like the arrow function `run: async (interaction) => {` is unnecessary, when you could just do `async run(interaction) {`

Now better?

Dragonizedpizza:
> > I think the snake case in the reroll command is unnecessary, considering that this is for the users to view, and Id looks weird. Also, the options' names must be lowercase (Discord API requires so), and I feel like the arrow function `run: async (interaction) => {` is unnecessary, when you could just do `async run(interaction) {`
> 
> Now better?

Yes, but why is the command run function async even though you're using .then?

Dragonizedpizza:
use async/await, its cleaner

Dragonizedpizza:
```js
await client.giveawaysManager.start(interaction.channel, { /** */ }),catch(() => { /** */ });
interaction.reply(yes)
```

Nico105:
> ```js
> await client.giveawaysManager.start(interaction.channel, { /** */ }),catch(() => { /** */ });
> interaction.reply("yes")
> ```
> 
> also, I think the replying should be deferred incase for whatever reason it takes more than 3 seconds.

applied, tho i went with try catch cause returning would have been a little "fancy" otherwise

src/Giveaway.js
```diff
@@ -543,7 +543,7 @@ class Giveaway extends EventEmitter {
                         this.messages.winMessage
                             .substr(0, this.messages.winMessage.indexOf('{winners}'))
                             .replace('{prize}', this.prize)
-                            .replace('{messageURL}', this.messageURL)
+                            .replace('{messageURL}', this.messageURL),
```
Nico105, line 546:
```suggestion
                            .replace('{messageURL}', this.messageURL)
```

Androz2091 (commented):
Thank you for this very huge PR. I guess that the best way is to keep minimal examples (I will make a commit don't worry), but the best would be to create an example bot like the one in the examples folder of discord-player. I will make it too, based on this PR. Thank you very much again, sorry this close this.

Issue #338, opened 2021-07-28
v13 installation & query


Dragonizedpizza:
When installing, I get this: ![image](https://user-images.githubusercontent.com/70718540/127337389-b34b397f-63eb-4379-8a8b-9323bf6774e5.png)/

Also, does Giveaway#messageID become Giveaway#messageId?

Nico105:
`npm i Androz2091/discord-giveaways#discord.js-v13`

Nico105:
And concerning ID to Id...
@Androz2091 you want to change?
Im not sure since it would be an unnecessary, possible breaking change, but on the other hand discord.js makes the transition, so maybe we should adjust too?

Dragonizedpizza:
> `npm i Androz2091/discord-giveaways#discord.js-v13`

oh wait, I'm dumb lol

Dragonizedpizza:
![image](https://user-images.githubusercontent.com/70718540/127339913-ce723682-9a30-440e-86c6-a36d792c2375.png)
oh yeah, this is what I was encountering

Nico105:
maybe `npm i git://github.com/Androz2091/discord-giveaways.git#discord,js-v13`
this seem to be a issue of the place wherever you try installing

Dragonizedpizza:
zsh's issue

Dragonizedpizza:
nvm query still stands

Dragonizedpizza:
uh...?

Dragonizedpizza:
messageId or messageID?

Nico105:
you will have to wait till Androz answers, since I don't want to charge forth.
I already put it into the v13 pr questions.

Androz2091:
Yes we should now use Id instead of ID, as discord.js v13 does

Nico105:
> Yes we should now use Id instead of ID, as discord.js v13 does

ok, and so since this breaks things 😆
suggestion to rename `time` to `duration` for the start() options
?

Nico105:
> Yes we should now use Id instead of ID, as discord.js v13 does

should `winnerIDs` stay or `winnerIds`?

Dragonizedpizza:
obviously winnerIds

Nico105:
yeah right `winnerIdentifications` is proper camelCase.
I just thought it looked a little weird to have Ids 😆

Dragonizedpizza:
ok you thought winnerIds looked weird and winnerIdentifications didn't

Nico105:
[...]

PR #362, opened 2021-08-23
v5.0.0

Androz2091:
PR for v5.0.0

TODO before merging:

- [x] Finish changelog
- [x] Update example bot at Androz2091/giveaways-bot with **all the features**
- [x] Test everything with the example bot
- [ ] Wait for @Nico105's review

Feel free to comment if you have anything else to add before it's merged


Lebyy:
Will #358 also be implemented in this release?

README.md
```diff
@@ -273,9 +324,17 @@ client.giveawaysManager.start(message.channel, {
...
 ```
 
+**Note**: because of the special `new Function()` format, you can use `this` inside of the function string to access anything from the giveaway instance. For example: `this.extraData`, or `this.client`.
+
+⚠️ **Note**: because of the special `new Function()` format, external variables which contain [Discord Snowflakes](https://discord.com/developers/docs/reference#snowflakes) have to be stored in `giveaway.extraData` and have to be accessed with `this.extraData` inside of the function.  
+The reason for this is that, in JavaScript, all numbers are [IEEE double precision](https://en.wikipedia.org/wiki/Double-precision_floating-point_format) floating point numbers, which means that you only have about 16 digits of precision; the remainder of the 64 bits are reserved for the exponent.  
+But snowflakes can have more than 16 digits, which means that if you would access them like in the example above, the last digits of the snowflake would get messed up.  
+_The special format would "convert" the external string snowflake into a number, that is why we talk about digits here._
+
+
```
Nico105, line 337: Is this somehow understandable? #ratemyenglish

Androz2091: yes, at least I understand. I will try to rewrite some parts of the README before the end of the day

Androz2091:
TODO before merging:

- [ ] Finish changelog
- [ ] Update example bot at Androz2091/giveaways-bot with **all the features**
- [ ] Test everything with the example bot
- [ ] Wait for @Nico105's review

Feel free to comment if you have anything else to add before it's merged

Lebyy:
> TODO before merging:
> 
> * [ ]  Finish changelog
> * [ ]  Update example bot at Androz2091/giveaways-bot with **all the features**
> * [ ]  Test everything with the example bot
> * [ ]  Wait for @Nico105's review
> 
> Feel free to comment if you have anything else to add before it's merged

> Will #358 also be implemented in this release?
^^

Androz2091:
Yes, it has already been merged @Lebyy :wink:

Androz2091:
@Nico105 should we rename `time` property to `duration`?

Androz2091:
and what is the point of the GUILD_MESSAGES intent?

Lebyy:
> Yes, it has already been merged @Lebyy :wink:

Oh! That's great :)

Nico105:
> @Nico105 should we rename `time` property to `duration`?

already asked you that in #338 😂
yeah i would say yes

Nico105:
> and what is the point of the GUILD_MESSAGES intent?

ah lol, i just searched. I whysoever had in mind that it was needed for interactions.
My bad.

Androz2091:
> already asked you that in #338 😂
> yeah i would say yes


Oops, looks like I missed your message 😅 
I will make the change then

Androz2091:
> ah lol, i just searched. I whysoever had in mind that it was needed for interactions.
> My bad.

This is required to receive user messages, that's not really needed. I will make this change too, then

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "04:00"
+  open-pull-requests-limit: 10
+  target-branch: develop
+  ignore:
+  - dependency-name: "@types/node"
+    versions: "^17.0.0"
```
Nico105, line 12: _idk if you oversaw it in the pr_
but would it be better if we just stick to 16.6.0 (= remove the "versions" option)?
since >.6. adds new things. but we can't/should't use them/the typings since our min node is only .6., right?

tsconfig.json
```diff
@@ -1,21 +1,18 @@
...
-        "esnext.asynciterable",
-        "esnext.intl",
-        "esnext.symbol"
-      ],
-      "sourceMap": false,
-      "skipDefaultLibCheck": true
-    }
-  }
\ No newline at end of file
+        "target": "es2020",
```
Nico105, line 3: i'm not sure if this is different for ts,
but we for example use replaceAll = 21 feature.
So should this get changed?

Androz2091: we can, but it does not matter as ts is only for typings

Nico105: ah, okay, then nvm I guess

Lebyy:
I suggest changing the screenshots in the README file to slash commands.

Nico105:
> I suggest changing the screenshots in the README file to slash commands.

i guess when we finalized slash usage.
And as mentioned in #358 , the other ss would technically also need an update to be correct,
But that comes after everything is finalized i guess.

Lebyy:
> > I suggest changing the screenshots in the README file to slash commands.
> 
> i guess when we finalized slash usage.
> And as mentioned in #358 , the other ss would technically also need an update to be correct,
> But that comes after everything is finalized i guess.

Yep, alright.

Androz2091:
I am building https://github.com/Androz2091/discord-giveaways, I will use it to make the screenshots

Androz2091:
https://github.com/Androz2091/giveaways-bot sorry 😅 
it did not mean anything

src/Manager.js
```diff
@@ -471,10 +471,12 @@ class GiveawaysManager extends EventEmitter {
...
-                giveaway.message.reactions.cache.get(giveaway.reaction)?.count - 1 >= giveaway.winnerCount
-            ) return this.end(giveaway.messageId).catch(() => {});
+            if (giveaway.isDrop) {
+                giveaway.message ??= await giveaway.fetchMessage().catch(() => {});
+                const reaction = giveaway.message?.reactions.find((r) => r.emoji.name === Discord.Util.resolvePartialEmoji(giveaway.reaction)?.name) ||
+                    giveaway.message?.reactions.get(Discord.Util.resolvePartialEmoji(giveaway.reaction)?.id);
+                if (reaction?.count - 1 >= giveaway.winnerCount) return this.end(giveaway.messageId).catch(() => {});
+            }
 
             // third case: the giveawat is paused and we should check whether it should be unpaused
```
Lebyy, line 481: There is a typo here in the comment. It says "giveawat" instead of "giveaway"

Lebyy (changes requested):

Dragonizedpizza:
Instead of String|null you can just do ?String

Lebyy:
Regarding the docs, do you have any plan to use [Yukitoki](yukitoki.snowflakedev.cf) as the renderer? Will look good, stylish and easy to read tbh.

Lebyy (dismissed):

Androz2091:
> Regarding the docs, do you have any plan to use Yukitoki as the renderer? Will look good, stylish and easy to read tbh.

We will think about that in the next version, for the moment let's fix the remaining bugs and release v5 👍

Lebyy:
> > Regarding the docs, do you have any plan to use Yukitoki as the renderer? Will look good, stylish and easy to read tbh.
> 
> We will think about that in the next version, for the moment let's fix the remaining bugs and release v5 👍

Alright.

src/Manager.js
```diff
@@ -181,85 +205,139 @@ class GiveawaysManager extends EventEmitter {
...
+            if (options.isDrop && typeof options.isDrop !== 'boolean') {
+                return reject(`options.isDrop is not a boolean. (val=${options.isDrop})`);
             }
-            if (typeof options.prize !== 'string') {
-                return reject(`options.prize is not a string. (val=${options.prize})`);
+            if (!options.isDrop && (isNaN(options.duration) || typeof options.duration !== 'number' || options.duration < 1)) {
+                return reject(`options.duration is not a positive number. (val=${options.duration})`);
             }
-            if (!Number.isInteger(options.winnerCount) || options.winnerCount < 1) {
+            if (!options.isDrop && (!Number.isInteger(options.winnerCount) || options.winnerCount < 1)) {
```
Nico105, line 228: this would currently result in an never ending drop. or how was this bypass intended?

src/Manager.js
```diff
@@ -181,85 +205,139 @@ class GiveawaysManager extends EventEmitter {
...
+                return reject(`options.prize is not a string or longer than 256 characters. (val=${options.prize})`);
             }
-            if (!options.time || isNaN(options.time)) {
-                return reject(`options.time is not a number. (val=${options.time})`);
+            if (options.isDrop && typeof options.isDrop !== 'boolean') {
+                return reject(`options.isDrop is not a boolean. (val=${options.isDrop})`);
             }
-            if (typeof options.prize !== 'string') {
-                return reject(`options.prize is not a string. (val=${options.prize})`);
+            if (!options.isDrop && (isNaN(options.duration) || typeof options.duration !== 'number' || options.duration < 1)) {
```
Nico105, line 225: _i am writing this with a extension, thats why no multiline, sry_
would it maybe be better to set endAt to Infinity? like for pausing? cause currently it would just be NaN

src/Manager.js
```diff
@@ -48,32 +52,61 @@ class GiveawaysManager extends EventEmitter {
...
+            )
             .setDescription(
-                (lastChanceEnabled ? giveaway.lastChance.content + '\n\n' : '') +
-                giveaway.messages.inviteToParticipate +
-                    '\n' +
-                    giveaway.remainingTimeText +
-                    '\n' +
-                    (giveaway.hostedBy ? giveaway.messages.hostedBy.replace('{user}', giveaway.hostedBy) : '')
+                giveaway.isDrop
+                ? giveaway.messages.dropMessage
```
Nico105, line 87: regarding that. if a user sets a duration, would it maybe be better to also display it?
= no duration => not in embed
= duration => in embed too

= switch behaviour
?

src/Constants.js
```diff
@@ -1,90 +1,95 @@
...
+ * @property {string} [hostedBy='Hosted by: {this.hostedBy}'] Below the "inviteToParticipate" message, in the description of the embed.
  */
-exports.GiveawayMessages = {};
+exports.GiveawayMessages = {
+    giveaway: '🎉🎉 **GIVEAWAY** 🎉🎉',
+    giveawayEnded: '🎉🎉 **GIVEAWAY ENDED** 🎉🎉',
+    inviteToParticipate: 'React with 🎉 to participate!',
+    winMessage: 'Congratulations, {winners}! You won **{this.prize}**!\n{this.messageURL}',
+    drawing: 'Drawing: {timestamp}',
+    dropMessage: 'Be the first to react with 🎉 !',
```
Nico105, line 27: would it maybe be a better idea to make a `dropOptions` object? like for pause and lastChance?
so that the boolean and the message are in 1 place.
and also cause it's more like a "external" feature, splitted from the normal gaws?

Nico105: Or on the other side, could it be it gets not displayed, because there is the timestamp?
but some people maybe oversee that, hmm, idk?

src/Manager.js
Nico105: Or on the other side, could it be it gets not displayed, because there is the timestamp?
but some people maybe oversee that, hmm, idk?

src/Constants.js
Nico105: + it would "save" space in the storage

Nico105 (commented):
image
_i will just add this for the record_

src/Manager.js
Androz2091: my bad, forgot to remove this

src/Manager.js
Androz2091: I added a default value for endAt for drop giveaways 👍

src/Manager.js
Androz2091: for me, a drop giveaway does not have any duration. Otherwise, users can still do it with the event when someone enters a giveaway and decide to end it if there is enough participations

src/Constants.js
Androz2091: there is only one extra prop (there is many other props with pause/last chance) so I think we can let it here, it does not really matter

Androz2091:
@Nico105 concerning your screenshots, I think that we must tell to users to use extra `'` when mapping the arrays. A snowflake should always be stored as a string

Nico105:
> @Nico105 concerning your screenshots, I think that we must tell to users to use extra `'` when mapping the arrays. A snowflake should always be stored as a string

`[${blarray.map((e) => `'${e.id}'`)}]`
yeah that works, finally.

src/Giveaway.js
```diff
@@ -441,155 +492,341 @@ class Giveaway extends EventEmitter {
...
+            if (!options || typeof options !== 'object') return reject(`"options" is not an object (val=${options})`);
+            options = merge(GiveawayRerollOptions, options);
+            if (options.winnerCount && (!Number.isInteger(options.winnerCount) || options.winnerCount < 1)) {
+                return reject(`options.winnerCount is not a positive integer. (val=${options.winnerCount})`);
             }
-            await this.fetchMessage().catch(() => {});
-            if (!this.message) {
-                return reject('Unable to fetch message with ID ' + this.messageID + '.');
+            if (this.isDrop) {
+                return reject(`Cannot reroll a drop giveaway.`);
```
Nico105, line 666: is this just a preference of yours?
cause it code wise it would be possible right?
so why not allow it? it doesn't have to be used after all

Androz2091: the goal of a drop is to be the fastest person to react. Why would we reroll it? it would be random, then it's not a drop anymore

Nico105: I guess that's true.
I just thought if somewhy a host considers a winner invalid.
But yeah if he rerolls it will be a random person. = not fair i guess
alright

Androz2091: if I were a server admin, I would instead create a new giveaway

src/Constants.js
Nico105: alright

Androz2091:
> But bruh, mid devs (like me XD) likely won't come to that conclusion 🥴

as I said it's an edge-case so if users have questions or errors about that they can come in our discord server

src/Constants.js
Nico105: Other thing: should the default message be adjusted to work for multiple winners too?
Cause currently it's "be the first", which might imply, by default, that there is only 1 winner.
_if you oversee the footer_

Androz2091: Drops are 99% of the time 1 winner. Like you're in a spammed chat and the first to react gains the nitro. And it's easier to understand. People can still check the docs to see if they can allow multiple winners

Nico105: alright

Nico105 (approved):
LGTM 👌

Nico105 (approved):

PR #366, opened 2021-09-03
Make embedEqual more efficient.

Dragonizedpizza:
## Changes

The old embedEqual function seemed large and inefficient, so I made it slightly smaller.

image

## Status

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,5 @@
-const Discord = require('discord.js');
+const Discord = require('discord.js'),
+    { serialize } = require('v8'); 
 
 const validateEmbedColor = (embedColor) => {
     try {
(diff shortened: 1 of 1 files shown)


Nico105:
image
The things is just that the embed from the message and the constructed embed are slightly different by default already.
So serialization wouldn't work.
Anyway, I don't think the current way is really inefficient. since it's just if statements and nothing spectacular.

Dragonizedpizza:
@Nico105 maybe I can remove all fields with undefined in them

Androz2091:
if you really see a case where the current function does not work, then feel free to PR otherwise I think that checking fields one by one is still the simplest solution (even simpler than a for loop..)

Dragonizedpizza:
> if you really see a case where the current function does not work, then feel free to PR otherwise I think that checking fields one by one is still the simplest solution (even simpler than a for loop..)

it just makes the code look less ugly

PR #404, opened 2021-12-16
✏️Use array fill instead of looping over the entries

Lebyy:
## Changes
This pr uses the js Array and Array fill method to multiple the user x many times instead of looping over the array.
The purpose of this pr is to solve the heap errors if the number is VERY LARGE thrown by node.

## Status

### DIFF
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -471,7 +471,7 @@ class Giveaway extends EventEmitter {
                 const highestBonusEntries = await this.checkBonusEntries(user);
                 if (!highestBonusEntries) continue;
                 
-                userArray = userArray.concat(Array(highestBonusEntries).fill(user));
+                userArray.push(...Array(highestBonusEntries).fill(user))
             }
         }
 
--- a/src/Giveaway.js
+++ b/src/Giveaway.js
@@ -470,8 +470,8 @@ class Giveaway extends EventEmitter {
 
                 const highestBonusEntries = await this.checkBonusEntries(user);
                 if (!highestBonusEntries) continue;
-
-                for (let i = 0; i < highestBonusEntries; i++) userArray.push(user);
+                
+                userArray = userArray.concat(Array(highestBonusEntries).fill(user))
             }
         }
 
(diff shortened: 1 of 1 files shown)


Dragonizedpizza:
you don't need to change the variable's value and concat the array, you could just do `userArray.push(...Array(highestBonusEntries).fill(user))`

Lebyy:
> userArray.push(...Array(highestBonusEntries).fill(user))

Oh yes, added that

Nico105:
image
about how large numbers are we talking about?
And can you explain a little more/proof the better efficiency, cause _ik its not to pratical but_ a little benchmarking and googling didn't really bring results...

Lebyy:
> image about how large numbers are we talking about? And can you explain a little more/proof the better efficiency, cause _ik its not to pratical but_ a little benchmarking and googling didn't really bring results...

Like 20k.

Nico105:
Now, I tested it with a bot and 2 entrants, 1 million bE `bonus: (member) => true ? 1_000_000 : null` and it was actually fill who got `Maximum call stack size exceeded` whereas the loop got it through...

Dragonizedpizza:
after some googling, I found this https://stackoverflow.com/a/53029824

The current way of doing it is the fastest.

Lebyy:
> after some googling, I found this https://stackoverflow.com/a/53029824
> 
> The current way of doing it is the fastest.

So we should use `let a = new Array(n); for (let i=0; i<n; ++i) a[i] = 0;` I think.

Nico105:
> let a = new Array(n)

image
normal push is still faster.
and going to 1mill, the 2nd one gets `Maximum call stack size exceeded`.

Dragonizedpizza:
@Nico105 what are you using for the benchmarking?

Nico105:
> @Nico105 what are you using for the benchmarking?

https://jsbench.me

Nico105:
> > let a = new Array(n)
> 
> image normal push is still faster. and going to 1mill, the 2nd one gets `Maximum call stack size exceeded`.

Any comments? If not then I would close this, unless...

PR #405, opened 2021-12-18
code refactored

imranbarbhuiya:
## Changes

- Embed timestamp accepts number as the parameter, so we don't need to convert `giveaway.endAt` to date 
- In many places, `!isNaN(input) && typeof input === 'number'` is used to validate the input is a number, but we can check that with `!Number.isNan(input)`

And sorry, that my prettier changed the code style.

## Status

### DIFF
--- a/src/Manager.js
+++ b/src/Manager.js
@@ -66,45 +66,22 @@ class GiveawaysManager extends EventEmitter {
     generateMainEmbed(giveaway, lastChanceEnabled = false) {
         const embed = new Discord.MessageEmbed()
             .setTitle(giveaway.prize)
-            .setColor(
-                giveaway.pauseOptions.isPaused && giveaway.pauseOptions.embedColor
-                    ? giveaway.pauseOptions.embedColor
-                    : lastChanceEnabled
-                        ? giveaway.lastChance.embedColor
-                        : giveaway.embedColor
-            )
+            .setColor(giveaway.pauseOptions.isPaused && giveaway.pauseOptions.embedColor ? giveaway.pauseOptions.embedColor : lastChanceEnabled ? giveaway.lastChance.embedColor : giveaway.embedColor)
             .setFooter(
-                typeof giveaway.messages.embedFooter === 'object'
-                    ? giveaway.messages.embedFooter.text?.length > 0
-                        ? giveaway.messages.embedFooter.text
-                        : ''
-                    : giveaway.messages.embedFooter,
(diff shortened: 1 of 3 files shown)


src/Manager.js
```diff
@@ -126,19 +103,15 @@ class GiveawaysManager extends EventEmitter {
 
         const descriptionString = (formattedWinners) => strings.winners + ' ' + formattedWinners + (giveaway.hostedBy ? '\n' + strings.hostedBy : '');
 
-        for (
-            let i = 1;
-            descriptionString(formattedWinners).length > 4096 ||
-            strings.prize.length + strings.endedAt.length + descriptionString(formattedWinners).length > 6000;
-            i++
-        ) formattedWinners = formattedWinners.substr(0, formattedWinners.lastIndexOf(', <@')) + `, ${i} more`;
```
imranbarbhuiya, line 134: Also changed substr to slice as substr is deprecated.

Dragonizedpizza:
Number.isNaN returns true incase the input is NaN.

imranbarbhuiya:
> Number.isNaN returns true incase the input is NaN.

Yeah that's why used `!`.

You r saying that if we give a input which isn't a Number then it returns true or you means anything else?

Nico105:
> > Number.isNaN returns true incase the input is NaN.
> 
> Yeah that's why used `!`.
> 
> You r saying that if we give a input which isn't a Number then it returns true or you means anything else?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#using_isnan
You see the problem... You would only be checking for NaN and nothing else

Nico105:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite
I stumbled across this. I think it would be most suitable, no?

imranbarbhuiya:
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite I stumbled across this. I think it would be most suitable, no?

I was unaware of this xD. Lmk should I change it to isFinite? As i think using a Number method instead checking isNan as well as checking number type will be better. 

Also I changed code from main branch am I supposed to change dobit from dev branch? If so I need to recreate the pr

Nico105:
> 

yeah change it to Number.isFinite for now, unless we find a exeption.
and for the branching, it should be fine if you edited the main branch since there are not to many differences.
But since you're prettier changed the formatting, it would be nice if you recreate/undo the prettier changes and just commit the real changes again. And then you could also start from the develop branch if you're already at it.

imranbarbhuiya:
Ok sure I'm closing this for now and will be recreating from dev branch to get rid of the conflict.

PR #412, opened 2021-12-28
refactor: import timers, event, from node:

imranbarbhuiya:
## Changes

- [x] I changed the fillInString and Embed type to return null only when the input isn't a string or it's undefined.
- [x] changed global setTimeout, setInterval, EventEmiter to their respective `node:*`  following https://github.com/discordjs/discord.js/pull/7157
- [x] allow users to specify extraData type

## Status

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -160,10 +160,10 @@ declare module 'discord-giveaways' {
         private ensureEndTimeout(): void;
         private checkWinnerEntry(user: User): Promise<boolean>;
         public checkBonusEntries(user: User): Promise<number>;
-        public fillInString(string: undefined | null): null;
         public fillInString(string: string): string;
-        public fillInEmbed(embed: undefined | null): null;
+        public fillInString(string: any): string | null;
         public fillInEmbed(embed: MessageEmbed | MessageEmbedOptions): MessageEmbed;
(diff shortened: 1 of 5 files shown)


typings/index.d.ts
```diff
@@ -170,8 +160,10 @@ declare module 'discord-giveaways' {
         private ensureEndTimeout(): void;
         private checkWinnerEntry(user: User): Promise<boolean>;
         public checkBonusEntries(user: User): Promise<number>;
-        public fillInString(string: string): string | null;
-        public fillInEmbed(embed: MessageEmbed | MessageEmbedOptions): MessageEmbed | null;
+        public fillInString(string: undefined | null): null;
```
Nico105, line 163: Maybe using a conditional type would be better, cause this doesn't include number, boolean etc. right?
And google didn't really tell me a way to specify "!string".

imranbarbhuiya: I've changed it, as u said. 

The output is now `string | null` it's because if you enter a data which is `string | undefined`, then it should return `string | null`. But if you enter a data which is string, then it'll return string only.

imranbarbhuiya (commented):
I turned off my prettier but my ts organize import changed it 😅.

Nico105 (approved):

PR #413, opened 2022-01-02
added extraData generic type to manager

imranbarbhuiya:
## Changes

- [x] added extraData generic type to manager so that it can give autocomplete in gw start options, edit and all. x-ref #412 
- [x] also change declare module to top level d.ts with exports
- [x] discord.js is now a dev dependency. Also added v13.0.0 in peer deps.

I did this because every user will be downloading djs, so there is no need to redownload the same.

## Status

## How to use these generic types?

You can use ExtraData interface in GiveawayManager's generic type. It's optional, if u don't provide the type it'll become `any` type as before.
```ts
  const manager = new GiveawayManager<ExtraData>(client, {
    default: {
      botsCanWin: false,
      embedColor: "#03FEE0",
      embedColorEnd: "#ffffff",
      reaction: "🎉",
    },
  });
```

Now you'll get autocomplete and validation in giveaway start options also.
image

If you are extending the manager to change generate embed, then you can add the extraData generic type in `Giveaway` class

```ts
generateMainEmbed(
    giveaway: Giveaway<GiveawayExtraData>,
    lastChanceEnabled = false
  ): MessageEmbed {
}
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,6 +19,29 @@
         "accessor-pairs": "off",
         "no-async-promise-executor": "off",
         "no-unused-vars": "off",
-        "node/no-missing-require": "off"
+        "node/no-missing-require": "off",
+        "no-restricted-globals": [
+            "error",
+            {
+              "name": "Buffer",
+              "message": "Import Buffer from `node:buffer` instead"
+            },
+            {
+              "name": "process",
+              "message": "Import process from `node:process` instead"
+            },
+            {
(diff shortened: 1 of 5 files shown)


imranbarbhuiya:
I accidentally edited in my last pr clone. Becouse of that all my last pr commits are also present here, idk how I can remove them now without recreating. Should I `recreate or squash` and merge will work?

imranbarbhuiya:
I'm going to recreate

PR #422, opened 2022-01-25
feat: add `MessageActionRow` components in MessageObject

imranbarbhuiya:
## Changes

This pr adds a new option for `components` in `MessageObject`

## Status


Nico105:
_finger slipped_

imranbarbhuiya:
> Also maybe filling in `linkURL` _or what its name is*_ would be good if someone wants to link to the gaw without editing the message after start()

I've changed the code to allow linkbutton and selectmenu.

imranbarbhuiya:
I'm not that good at wording. Please change it to whatever will suit.

Nico105:
both blocks are referring to `winMessage`...
But anyway, we would need such a block for noWinnerMessage and error of rerroll too.
So lets just make it a `fillInComponents` function, no?

imranbarbhuiya:
> both blocks are referring to `winMessage`...
> But anyway, we would need such a block for noWinnerMessage and error of rerroll too.
> So lets just make it a `fillInComponents` function, no?

I've updated the code to allow no winner messages also.

src/Giveaway.js
```diff
@@ -371,6 +371,30 @@ class Giveaway extends EventEmitter {
         return embed;
     }
 
+    /**
+     * @param {Discord.MessageActionRow[]} components The components that should get filled in.
+     * @returns {Discord.MessageActionRow[]} The filled in components.
+     */
+    fillInComponents(components) {
+        if (!components?.length) return components;
```
Nico105, line 379: wouldn't returning null like the for the other methods be better?
cause otherwise the returned value might not match the doc

Nico105: [MessageOptions](https://github.com/discordjs/discord.js/blob/988a51b7641f8b33cc9387664605ddc02134859d/src/structures/MessagePayload.js#L141)
they don't explicitly state it, but it actually is nullable, same in v14 dev

imranbarbhuiya: fixed

Nico105:
we would have to add [`| MessageActionRowOptions`](https://discord.js.org/#/docs/discord.js/stable/typedef/MessageActionRowOptions) too right?
Cause it's at least guaranteed through serialization in case of winMessage.

imranbarbhuiya:
> we would have to add [`| MessageActionRowOptions`](https://discord.js.org/#/docs/discord.js/stable/typedef/MessageActionRowOptions) too right? Cause it's at least guaranteed through serialization in case of winMessage. (function behaviour doesn't change)

added

Nico105 (approved):
should be fine right?

imranbarbhuiya:
> should be fine right?

Yeah.

Nico105:
👍

PR #435, opened 2022-02-07
Consistent variable declaration syntax.

Dragonizedpizza:
## Changes

 The README uses commas while declaring variables, to declare multiple variables at the top, but it isn't continued later on.

## Status

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,26 +35,26 @@ Required Discord Intents: `GUILDS` and `GUILD_MESSAGE_REACTIONS`.
 Optional Discord Privileged Intent for better performance: `GUILD_MEMBERS`.
 
 ```js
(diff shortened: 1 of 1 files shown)


Nico105:
the manager should be left standing out, for better distinction.
If a change should be made, then to all const imo.

Cause we can remove the `prefix` cause our examples are for slash anways now.
And remove token and do `client.login(process.env.DISCORD_BOT_TOKEN)` instead.

Dragonizedpizza:
How about no?

imranbarbhuiya:
Imo the const related change isn't needed. You are just saving 4 letters. But writing const twice is much ux friendly and I prefer it for readability.

PR #436, opened 2022-02-08
feat: add option to remove reroll error message 

### LINKED ISSUE #431: Stop failed reroll message, customize end message as a Message
Firstly, when I reroll a giveaway and there aren't enough participants, this message is sent in the chat.
image
Can I stop this from happening?

Secondly, can I modify the giveaway end message as a Message? Like add components to it.

imranbarbhuiya:
## Changes

This PR adds a new reroll option to remove error message 
closes #431 
## Status


imranbarbhuiya:
idk why with the same prettier config my prettier formats are different than yours. @Nico105 please format again with your prettier.

src/Giveaway.js
```diff
@@ -900,18 +900,24 @@ class Giveaway extends EventEmitter {
...
-                    allowedMentions: this.allowedMentions,
-                    reply: {
-                        messageReference:
-                            typeof options.messages.error.replyToGiveaway === 'boolean' ? this.messageId : undefined,
-                        failIfNotExists: false
-                    }
-                });
-                resolve([]);
+                if (options.rejectOnNoWinner) {
+                    reject(new Error('No valid participations, no new winner(s) can be chosen!'));
```
imranbarbhuiya, line 904: here i've used a `Error` instance but I just noticed In all other places it's `reject(reason)`. I prefer using `Error` if u want me to change in all other places to `Error` or revert this one to only reason then lmk.

imranbarbhuiya:
This can be added in `end` also, but should I make it the 3rd param in `manager#end` or make options an object?

Nico105:
> This can be added in `end` also, but should I make it the 3rd param in `manager#end` or make options an object?

make it an object.

Nico105:
Wouldn't naming it `rejectWhenNoWinner` "sound" more english?

imranbarbhuiya:
> Or maybe don't go with the rejecting at all and just add a `active: true/false` option to the MessageObject option of the `error` message. In order to make it easier to verifiy the process. Because there are other reject reasons, and to filter out this one you would have to `if` check the rejection or check the winnerIds array by fetching the gaw. disabling the msg and letting the method resolve would offer the convenience of being able to check the resolved winner array for easier handling.

Ic buyt currently

imranbarbhuiya:
Sorry wrong click

README.md
```diff
@@ -115,37 +116,45 @@ This allows you to start a new giveaway. Once the `start()` function is called,
...
+        (g) => g.guildId === interaction.guildId && g.prize === interaction.options.getString('query')
+    ) ||
+    // Search with messageId
+    client.giveawaysManager.giveaways.find(
+        (g) => g.guildId === interaction.guildId && g.messageId === interaction.options.getString('query')
+    );
 
 // If no giveaway was found
-if (!giveaway) return interaction.channel.send('Unable to find a giveaway for `'+ args.join(' ') +'`.');
+if (!giveaway) return interaction.channel.send('Unable to find a giveaway for `' + args.join(' ') + '`.');
```
imranbarbhuiya, line 134:
```suggestion
if (!giveaway) return interaction.reply('Unable to find a giveaway for `' + options.getString('query') + '`.');
```
In interaction, it's possible that the bot doesn't have permission to send messages in this channel. So changing it to reply will solve the issue.

Nico105: yeah you can

src/Giveaway.js
```diff
@@ -915,17 +915,21 @@ class Giveaway extends EventEmitter {
...
-                    content: this.fillInString(options.messages.error.content || options.messages.error),
-                    embeds: embed ? [embed] : null,
-                    allowedMentions: this.allowedMentions,
-                    reply: {
-                        messageReference:
-                            typeof options.messages.error.replyToGiveaway === 'boolean' ? this.messageId : undefined,
-                        failIfNotExists: false
-                    }
-                });
+                if (options.messages.replyWhenNoWinner) {
```
Nico105, line 918:
```suggestion
                if (options.messages.replyWhenNoWinner !== false) {
```
i think it's better to only deactivate with a real boolean

src/Constants.js
```diff
@@ -163,12 +163,14 @@ exports.GiveawaysManagerOptions = {
  * @property {Object} [messages] The messages used in this method.
  * @property {string|MessageObject} [messages.congrat=':tada: New winner(s): {winners}! Congratulations, you won **{this.prize}**!\n{this.messageURL}'] The message used if there are new winners.
  * @property {string|MessageObject} [messages.error='No valid participations, no new winner(s) can be chosen!'] The message used if no new winner(s) could be chosen.
+ * @property {boolean} [messages.replyWhenNoWinner=true] Whether or not reject the promise if there is no winner.
```
Nico105, line 166:
```suggestion
 * @property {boolean} [messages.replyWhenNoWinner=true] Whether or not to send the "error" message when there is no winner.
```

Nico105 (approved):
👍

PR #444, opened 2022-03-18
chore: export Giveaway class

imranbarbhuiya:
## Changes

Export Giveaway class.

Why?

- You can now extend `start` method where `Giveaway` class is needed.
- You can now use a custom `Manager` with the giveaway class.
- You can extend `Giveaway` class to do any changes and then use it in `GiveawayManager#start` to use extended `Giveaway` class.
## Status

### DIFF
--- a/esm.mjs
+++ b/esm.mjs
@@ -1,6 +1,3 @@
 import DiscordGiveaways from './index.js';
 
-export const {
-    GiveawaysManager,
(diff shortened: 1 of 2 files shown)


Nico105:
alright, but for the record I have to say: people who change too much in it will be on their own for bug fixing.

PR #448, opened 2022-03-22
Update Giveaway.js

typhoon11:
## Changes
Added catch in guild.members.fetch when intents are on.

## Status


Nico105:
It actually is added on dev already. you're base is different or smth

PR #452, opened 2022-04-06
feat: add Giveaway param in exemptmember

imranbarbhuiya:
## Changes

This pr adds `Giveaway` param in `exemptMembers` function. 

Ik we've `new Function()`, where we can access `this`, but when using `new Function()` all our javascript code becomes a string and that's why our ide/typescript can't show errors. Also, everyone doesn't know how to make `new Function()` async.
## Status

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,5 +1,5 @@
-import { EventEmitter } from 'node:events';
-import {
+import type { EventEmitter } from 'node:events';
+import type {
     Client,
     ColorResolvable,
     EmojiIdentifierResolvable,
@@ -16,16 +16,17 @@ import {
     Snowflake,
     TextChannel,
     ThreadChannel,
-    User
+    User,
+    Awaitable
 } from 'discord.js';
 
 export const version: string;
 export class GiveawaysManager<ExtraData = any> extends EventEmitter {
-    constructor(client: Client, options?: GiveawaysManagerOptions, init?: boolean);
+    constructor(client: Client, options?: GiveawaysManagerOptions<ExtraData>, init?: boolean);
 
     public client: Client;
     public giveaways: Giveaway<ExtraData>[];
-    public options: GiveawaysManagerOptions;
+    public options: GiveawaysManagerOptions<ExtraData>;
     public ready: boolean;
 
     public delete(messageId: Snowflake, doNotDeleteMessage?: boolean): Promise<Giveaway<ExtraData>>;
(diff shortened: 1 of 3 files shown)


typings/index.d.ts
```diff
@@ -180,15 +181,15 @@ export class Giveaway<ExtraData = any> extends EventEmitter {
     readonly isDrop: boolean;
 
     private ensureEndTimeout(): void;
-    private checkWinnerEntry(user: User): Promise<boolean>;
-    public checkBonusEntries(user: User): Promise<number>;
+    private checkWinnerEntry(user: User): Awaitable<boolean>;
+    public checkBonusEntries(user: User): Awaitable<number>;
```
Nico105, line 185: I'm not the most theoretical guy but just some speculation on my part.
These 2 functions should be awaited/use promises.
But the "Awaitable" is defined with an "OR" non-promise value.
Wouldn't that maybe imply wrong usage capability of the function for the develeoper? _Might be wrongly worded by me_

/ the whole "promise | non-promise" is imo just for the values the user provides for the properties. "= no execution"
-> but the return values of the async functions should be treated as a promise because of "excecution with promises".

As said just thoughts of mine. / if my interpretation is just my coding "convenience" then i'm sorry

Nico105:
btw for consistency you could also add the param for bonusEntry functions, if you want.

typings/index.d.ts
imranbarbhuiya: yeah I missed that thank you.

typings/index.d.ts
```diff
@@ -191,7 +192,7 @@ export class Giveaway<ExtraData = any> extends EventEmitter {
     public fillInEmbed(embed: unknown): MessageEmbed | null;
     public fillInComponents(components: (MessageActionRow | MessageActionRowOptions)[]): MessageActionRow[];
     public fillInComponents(components: unknown): MessageActionRow[] | null;
-    public exemptMembers(member: GuildMember): Promise<boolean>;
+    public exemptMembers(member: GuildMember, giveaway: Giveaway): Awaitable<boolean>;
```
Nico105:
```suggestion
    public exemptMembers(member: GuildMember): Promise<boolean>;
```
this one is the function which executes the exemptMembers function of the user. not the user function itself.

imranbarbhuiya: coincidently I fixed it 1s ago xD

imranbarbhuiya:
> btw for consistency you could also add the param for bonusEntry functions, if you want.

I haven't checked any bonus entry-related codes yet so idk how it works and I don't have time to check all the bonus entries related codes. If anyone else is interested the go ahead

Nico105:
> > btw for consistency you could also add the param for bonusEntry functions, if you want.
> 
> I haven't checked any bonus entry-related codes yet so idk how it works and I don't have time to check all the bonus entries related codes. If anyone else is interested then go-ahead

np. i will do it when i have time then.

PR #462, opened 2022-05-06
🐛 Check if embed was edited on end()

Nico105:
## Changes

Added private `isEnding` property which makes sure that `ended` is only true if the giveaway embed was successfully edited.

The only thing that isn't checked is wheter all win Messages were sent. But checking that would require too much fetching, so it's not worth it. It's most important that the embed is properly edited.

## Status

- [ ] These changes have been tested and formatted properly.


Nico105:
_Hope this goes well, otherwise revert and thinking of a new approach XD_

Nico105:
_didn't go well => will have reconsider the approach_

PR #478, opened 2022-08-22
fillInEmbed: fix data source of embed

kenmadev:
## Changes
Since we shifted using discordjs v14, `EmbedBuilder` uses a `data` property to access the properties of the embed. 
The `Discord.EmbedBuilder.from` method accepts the same parameter with `APIEmbed` [interface](https://discord-api-types.dev/api/discord-api-types-v10/interface/APIEmbed) but we're currently passing the embed instance to the said function, this results to have a redundant `data` property. Giving `{data: {data:{}}}` structure. It happens when we pull data from a database.

When a giveaway ended. The line https://github.com/Androz2091/discord-giveaways/blob/master/src/Giveaway.js#L751 throws an error. Because the description is null coming from https://github.com/Androz2091/discord-giveaways/blob/master/src/Giveaway.js#L745-L746. Again it only happens when pulling the giveaway from DB where the timer runs out but has not ended.

## Status

### DIFF
--- a/src/Constants.js
+++ b/src/Constants.js
@@ -229,6 +229,7 @@ exports.GiveawayEditOptions = {};
  * @property {Discord.Snowflake} guildId The Id of the guild.
  * @property {Discord.Snowflake[]} [winnerIds] The winner Ids of the giveaway after it ended.
  * @property {Discord.Snowflake} messageId The Id of the message.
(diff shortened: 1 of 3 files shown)


Tomato6966:
do `embed?.data ?? embed`

ppl can then still provide raw api richembed data

Nico105:
Can you provide a step by step reproduce instruction for both of these issues?
Cause I seem to do the wrong things for repro

PR #480, opened 2022-09-11
Adde entriesAmount variable for Giveaway when ended | update discord.js to 14.22.1 | update ready -> clientReady | add prisma example

Tomato6966:
## Changes
Added the variable entriesAmount to a Giveaway
so you can add this in an ending message:

```
{this.entriesAmount} Entries
```

image

`npm i Androz2091/discord-giveaways#9c44927bd696e9a06e494af9554d27c1718e997b`

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,9 +50,6 @@
     "jsdoc-skyceil": "Androz2091/jsdoc-skyceil",
     "typescript": "^4.7.4"
   },
-  "peerDependencies": {
-    "discord.js": ">=14.0.0"
-  },
   "engines": {
     "node": ">=16.9.0"
   }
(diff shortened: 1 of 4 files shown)


Nico105:
For more clarity, I think smth like `finalEntrantAmount` would go better.
and setting it to null instead of 0 seems more appropriate, no?
also, I guess this should be saved to the db?

Tomato6966:
Sry for late respond, it really is useful, as in most giveaway bots, they show the amount of entries after the giveaway ended, this way it's the most easiest to implement, just add the variable {this.entriesAmount} inside your template messages strings

Tomato6966:
@Nico105 Any updates?

PR #507, opened 2023-06-28
fix: add "boolean" type to storage

GlitchApotamus:
This allows typescript users to write ***storage: false*** in the options.

## Status

- [ x ] These changes have been tested and formatted properly.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -87,7 +87,7 @@ export interface PauseOptions {
     infiniteDurationText?: string;
 }
 export interface GiveawaysManagerOptions<ExtraData> {
-    storage?: string;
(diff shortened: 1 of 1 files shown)


Nico105:
If I'm not mistaken it's not necessary to set it to false anymore.
The code example with that included should be removed.
So you can just not use it / leave it undefined, no?

GlitchApotamus:
When I don't use it, it defaults to "./giveaways.json" and undefined will also fault out in typescript because it's not a string

Nico105:
But so are you using a custom db?
Because if you are extending the getAllGiveaways function then the "default" shouldn't matter to you.

So I don't really understand what you are trying to achieve/what your logic is.

PR #513, opened 2024-10-30
📝 Fix Unknow Interaction

Exodo0:
## Changes

I updated part of the code so that the Unknow Interactions would stop appearing. This is because if you use buttons. For example:

When joining the giveaway for some reason it showed: You are already joined to the giveaway. This was false since it was the first time the user joined.

I tested it and it works. Now it doesn't give an error and everything is registered correctly.

## Status

- [ X ] These changes have been tested and formatted properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,17 +38,17 @@
   },
   "homepage": "https://discord-giveaways.js.org",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


Androz2091:
@Nico105 could you review this if you have some time?

